PR #266, opened 2018-08-06
chore(deps): update js dependencies

mortik:


mortik:
Webpacker not yet ready for new vue-loader

PR #463, opened 2019-02-25
chore(e2e): add screenshot compare

mortik:


mortik:
Will revisit this later with a Cypress Plugin

PR #607, opened 2019-05-21
chore(precompile): use only rails tasks

mortik:


mortik:
Closed in favor of #609

PR #1432, opened 2021-04-15
chore: webpacker 6 upgrade

mortik:


mortik:
Closed because Webpacker 6 was deprecated. Topic will be further dealt with in #2099

PR #2441, opened 2022-10-31
feature: Basic Lists for saving Wishes and more

mortik:


mortik:
Currently obsolete with the new changes to how purchased works.

PR #2481, opened 2022-12-05
chore: Small Embed CSS Refactor

mortik:


mortik:
without scoping styles a leaking into the parent page... therefore for now closed

PR #2499, opened 2022-12-26
chore(api-v2): Implement OpenAPI based API

mortik:


mortik:
Will be implemented here https://github.com/fleetyards/ExFleetYards

PR #2512, opened 2023-01-11
Devenv

kloenk:
Nix devenv including sql server


mortik (approved):

kloenk:
The macOS nix build fails because of broken v8 in upstream. I would suggest either ignoring, or disabling macOS builds.

mortik:
> The macOS nix build fails because of broken v8 in upstream. I would suggest either ignoring, or disabling macOS builds.

Would be fine with me to only run nix build on ubuntu in the pipeline.

PR #2572, opened 2023-03-11
chore: migrate deployment and provisioning to pulumi an mrsk

mortik:


NargiT:
more will come for FR

mortik:
resolved by #3238

PR #3102, opened 2024-10-22
Fix ATLS Sync to Hanger

### LINKED ISSUE #3101: ATLS not picked up on RSI Sync
Whendoing an RSI Hanger Sync the ATLS comes back as A.T.L.S. thus it doesn't get matched to your sites entry of ATLS, this needs to ether be aliased or name changed to allow it to match up to the data returned from RSI

image

image

TheWhiteWolves:
Add missing dot in ATLS mapping to fix #3101 and correctly sync the vehicle from RSI API to Fleetyards Hanger

### DIFF
--- a/app/lib/hangar_sync.rb
+++ b/app/lib/hangar_sync.rb
@@ -364,7 +364,7 @@ class HangarSync < HangarImporter
     name = name.tr("–", "-")
 
     mapping = {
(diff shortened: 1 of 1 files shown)


TheWhiteWolves:
Fixed in 4e41ca304e5cd1a5c69d567ea105b2c43b5327d7

PR #4104, opened 2026-06-11
chore(test): Phase 2 batch 10 — admin models

mortik:
Phase 2 batch 10. Ports 15 admin model specs into 11 Minitest test classes. 54 tests / 115 assertions; `swagger/*` byte-identical.

Heavy splitting required for find_context_for: two GET endpoints without path params, five PUT endpoints without path params, and five PUT endpoints with path params all needed their own test class.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mortik:
Thanks — good catch. Holding off on the fixture move until all Phase 2 batches land on `main`. Doing it now would mean updating 11 references across factories shared with the still-live RSpec specs (and the same path appears in `images_test.rb` and `email_processor_test.rb` in this PR too), so a single sweep post-Phase-2 is cleaner. Tracking it as part of the Phase 3 spec/→test/ move. 🤖

PR #4109, opened 2026-06-11
chore(lint): apply standardrb 1.55 autofixes

mortik:
## Summary
- Pre-applies the lint changes required by standard 1.55 so #4095 can land cleanly.
- Removes redundant `keyword_init: true` from 4 Structs (Ruby 3.2+ accepts kwargs by default — and call sites already use kwargs).
- Adjusts indentation in two chained `.map`/`.find_each` block bodies per the new `Layout/IndentationWidth` interpretation.

## Test plan
- [x] `bundle exec standardrb` is clean locally with standard 1.55.0.
- [ ] CI green.

🤖


mortik:
Closing — the indentation autofix produced by standard 1.55 fails lint under 1.54, so it can't land standalone. Squashing the fixes into #4095 instead. 🤖

PR #4152, opened 2026-06-21
feat: notify the team when a new Patreon supporter joins

mortik:
## What

Sends a notification to the team whenever the Patreon sync imports a **brand-new active patron**, and hides the admin "Sync from Patreon" button when Patreon credentials aren't configured.

Builds on the Patreon supporter sync (#4150).

## Notifications

When a sync creates a new active patron, two alerts fire via `Notifications::NewPatronJob`:

- **Email** to super admins (`AdminMailer#new_supporter`) — always sends.
- **Discord webhook** (`Discord::NewSupporter`) to an internal channel — only sends if the new `discord_admin_endpoint` credential is set; otherwise a silent no-op.

Both include the supporter's name and pledge amount (e.g. `5.00 EUR`).

The trigger lives in `Patreon::SupporterImporter#upsert` and fires only for `new_record && ended_at.blank?` — so updates, reactivations, and members created already-churned do **not** notify.

## Hide sync button when unconfigured

- The admin stats endpoint now returns `patreonSyncEnabled` (`PatreonSupporterSyncJob.configured?`).
- The frontend gates the "Sync from Patreon" button on it, reusing the cached stats query the page header already fetches (no extra request).

## Config

The Discord half depends on a new credential; the email works without it:

```yaml
discord_admin_endpoint: <discord-webhook-url>
```

Documented in the README.

## Testing

- New tests: `NewPatronJob`, `Discord::NewSupporter`, importer enqueue behaviour, `AdminMailer#new_supporter`, and a stats-endpoint assertion for `patreonSyncEnabled`.
- All affected suites pass; `standardrb`, ESLint, Prettier and `vue-tsc` clean.

## Notes

- Used a dedicated internal Discord endpoint rather than the public `discord_updates_endpoint`, since synced patrons are anonymous by default and the message names them.
- The generated `app/frontend/services/` client is gitignored (rebuilt at deploy), so CI will regenerate the TS client from the updated schema.

🤖

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,16 @@ USD pledge amounts are converted to EUR at sync time using the free
 to the `exchange_rates` table as a fallback when the rate API is unreachable. No
 extra credentials are required for the exchange-rate lookup.
 
+When a sync creates a brand-new active patron, the team is notified by email
+(super admins) and via a Discord webhook. The Discord post is sent to an internal
+channel and is skipped unless its webhook is configured:
+
+```yaml
+discord_admin_endpoint: <discord-webhook-url>
+```
+
+The email always sends; only the Discord half depends on this credential.
+
 ## Authors and Acknowledgement
 
 - [@mortik](https://www.github.com/mortik) for development and design.
--- a/app/api_components/admin/v1/schemas/supporter_contribution_stats.rb
+++ b/app/api_components/admin/v1/schemas/supporter_contribution_stats.rb
@@ -13,10 +13,11 @@ module Admin
             currency: {type: :string},
             totalCount: {type: :integer},
             recurringCount: {type: :integer},
-            anonymousCount: {type: :integer}
+            anonymousCount: {type: :integer},
+            patreonSyncEnabled: {type: :boolean}
           },
           additionalProperties: false,
-          required: %w[totalAmountCents currency totalCount recurringCount anonymousCount]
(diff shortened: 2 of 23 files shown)


mortik:
Thanks for the review — addressed in `7db8d2084` and `a821bc19d`.

**Fixed:**
- **Discord failure breaking email delivery / duplicate posts on retry** (Greptile, Qodo) — the email is now enqueued *before* the Discord post, and the webhook call is wrapped in a `rescue` that reports to Appsignal. A transient Discord failure no longer aborts the job or re-fires the webhook on retry.
- **Blank patron name** (Greptile) — added `SupporterContribution#display_name` with a localized anonymous fallback, used by both the Discord message and the email.
- **Discord mention injection** (Qodo) — the webhook now sets `allowed_mentions = { parse: [] }`, so a patron name like `@everyone` can't ping the channel.

**Not changing, with reasoning:**
- **`# rubocop:disable Naming/AccessorMethodName`** — identical to the existing `Discord::NewShip` / `Discord::Webhook` directives; it's required because the base class's interface methods are named `get_title`/`get_message`/`get_url`. Consistent with the established pattern.
- **Single-quoted string literals** — this repo lints with **standardrb**, not RuboCop, and standardrb prefers double quotes (the whole codebase uses them). standardrb passes clean.
- **RSpec instead of Minitest** — the repo is migrating *to* Minitest; all tests here are Minitest by design (cf. #4075, #4147).

🤖

PR #4153, opened 2026-06-21
Log API request identity and close the rate-limit referer bypass

mortik:
## Why

While investigating whether the public API catalog (`/v1/models`) was being harvested, two gaps surfaced:

1. **No way to attribute API traffic.** Lograge ran with its stock config, so request logs carried no `remote_ip`, `user_agent` or `referer` — leaving no way to tell *who* was making bulk requests. AppSignal traces don't capture these either (and are sampled).
2. **The API rate limit had a spoofable bypass.** The `rack-attack` throttle exempted any request whose `Referer` started with `https://fleetyards.net`. `Referer` is a client-controlled header, so any caller could send that one line and obtain **unthrottled** access to the public API.

## What changed

- **`config/initializers/lograge.rb`** (new): add a `custom_payload` so every request log line includes `remote_ip`, `user_agent` and `referer`. `remote_ip` resolves the real client IP through the Traefik proxy. This makes catalog harvesting (e.g. `page=1&per_page=240` with no session) traceable to a caller.
- **`config/initializers/rack_attack.rb`**: drop the `Referer`-based exemption. All anonymous API traffic on the `api.*` subdomain is now throttled per IP; authenticated `api_user`s keep the higher 10k/hour limit via the existing `limit_proc`.

## Tradeoff to note

The removed exemption also covered *logged-out* first-party browsing on fleetyards.net, which now shares the **5,000/hour per-IP** anonymous bucket. That's generous for an individual user (~83 req/min), but visitors behind shared/CGNAT IPs (mobile carriers, universities, offices) now aggregate against one limit. If that proves too tight, the clean follow-up is to raise the anonymous limit in `limit_proc` rather than re-introduce a spoofable bypass.

## Verification

- Booted Rails and confirmed `custom_payload_method` emits `{remote_ip, user_agent, referer}`.
- Confirmed the `api` throttle still loads after removing the exemption.
- Both files pass `standardrb`. No existing rack-attack specs (the bypass was untested).

🤖

### DIFF
--- a/config/initializers/lograge.rb
+++ b/config/initializers/lograge.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+Rails.application.configure do
+  config.lograge.custom_payload do |controller|
+    {
+      remote_ip: controller.request.remote_ip,
+      user_agent: controller.request.user_agent,
+      referer: controller.request.referer
+    }
+  end
+end
--- a/config/initializers/lograge.rb
+++ b/config/initializers/lograge.rb
@@ -1,11 +1,14 @@
 # frozen_string_literal: true
 
 Rails.application.configure do
+  sanitize = ->(value) { value && value.to_s.delete("\r\n").truncate(512) }
+
   config.lograge.custom_payload do |controller|
+    request = controller.request
     {
-      remote_ip: controller.request.remote_ip,
-      user_agent: controller.request.user_agent,
-      referer: controller.request.referer
+      remote_ip: request.remote_ip,
+      user_agent: sanitize.call(request.user_agent),
+      referer: sanitize.call(request.referer&.split("?", 2)&.first)
     }
   end
 end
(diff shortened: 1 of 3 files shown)


mortik:
Addressed the automated review feedback:

**Fixed**
- **IP source mismatch** (Qodo) — the throttle now keys on the same `action_dispatch.remote_ip` resolution used for logging (falling back to `req.ip`), so the throttled IP matches what shows up in logs during an investigation.
- **Raw header logging** (Qodo, security) — the referer query string is now stripped (it can carry OAuth codes / secrets), CRLF is removed to prevent log injection, and both `user_agent` and `referer` are capped at 512 chars.

**Deferred to a follow-up (not this PR)**
- **MemoryStore counters are per-process** (Greptile) — valid and worth doing, but it's pre-existing and a broader operational change (making the throttle depend on Redis on every request). Tracking separately rather than expanding this PR's scope.

**Not actioned**
- **Single- vs double-quoted strings** (Qodo) — this repo lints with `standardrb`, which passes and doesn't enforce that rule; the line is also pre-existing.
- **Missing RSpec coverage** (Qodo) — happy to add a throttle spec if wanted, but rate-limit specs tend to be heavy/flaky and there's no existing rack-attack coverage to extend.

🤖

PR #4212, opened 2026-08-03
feat(api): type resourceAccess with dedicated enums

mortik:
## Why

`resourceAccess` was typed as a loose `string[]` across the API and generated frontend types, so nothing tied the values to the actual privilege catalogs and typos in `access:` checks went uncaught.

## What

Replace the loose `string[]` with three dedicated enums:

| Enum | Source of truth | Values |
|---|---|---|
| `FleetRoleResourceAccessEnum` | `FleetRole.all_available_privileges` (referenced live — zero drift) | 30 fleet privileges (`fleet:manage`, `fleet:roles:manage`, …) |
| `AdminUserResourceAccessEnum` | Admin route `access:` meta + `stats` | 18 admin privileges (`models`, `users`, `pghero`, …) |
| `UserResourceAccessEnum` | Empty — matches the always-`[]` serializer | none |

- Wired the enums into `FleetRole`, `User`, `AdminUser`, and `AdminUserInput` schema components via `$ref`.
- Regenerated the OpenAPI schema and Orval API clients (`./bin/generate-schema`).
- Fixed the 3 consumers the narrowed types surfaced — `.includes(dynamicString)` in both `session.ts` stores and `roles.vue` now widen the array to `string[]`, matching how `checkAccess` already types its param.

## Notes / follow-ups

- `AdminUserResourceAccessEnum` has **no backend source of truth** (admin privileges live only in frontend route meta), so it won't auto-update when a new admin resource is added. A follow-up could add an `AdminUser::AVAILABLE_PRIVILEGES` constant to make it drift-proof like FleetRole.
- `UserResourceAccessEnum` is intentionally empty — `User.resourceAccess` is serialized as `[]` with no consumers. The field is effectively dead and could be removed later.

## Verification

- `standardrb` — clean
- `vue-tsc` + `tsc` — clean
- `prettier` + `eslint` — clean

🤖

### DIFF
--- a/app/api_components/admin/v1/schemas/admin_user.rb
+++ b/app/api_components/admin/v1/schemas/admin_user.rb
@@ -15,7 +15,7 @@ module Admin
             twoFactorRequired: {type: :boolean},
             twoFactorQrCodeUrl: {type: :string},
             twoFactorProvisioningUrl: {type: :string},
-            resourceAccess: {type: :array, items: {type: :string}},
+            resourceAccess: {type: :array, items: {"$ref": "#/components/schemas/AdminUserResourceAccessEnum"}},
             superAdmin: {type: :boolean},
             createdAt: {type: :string, format: "date-time"},
             updatedAt: {type: :string, format: "date-time"}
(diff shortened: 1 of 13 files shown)


mortik:
Superseded by #4213 — same commits, continued on the `feat/resource-access-grouped-catalog` worktree branch where the grouped-catalog follow-up will land. 🤖

PR #4217, opened 2026-08-04
fix(paints): reuse open import issues and only raise on problems

mortik:
## What & why

The paints import mishandled its GitHub issues in two ways:

1. **Duplicate issues.** `GithubIssueCreator#run` always called `create_issue`, so every content change opened a **brand-new** issue instead of updating the one it had already logged — leaving a pile of stale duplicates.
2. **Noise on clean runs.** `PaintsImportJob` created an issue on every run, even when everything imported cleanly.

## Changes

- **`GithubIssueCreator`** now looks up the last logged issue for the `task_type`, checks whether it's still open on GitHub, and **updates that issue** rather than opening a new one. A fresh issue is only created when there's no open one to reuse (or the previous was closed/deleted). Since this is the shared creator, modules and loaner imports get the same reuse behavior.
- **`PaintsImportJob`** now only raises an issue when `new_with_error` or `model_not_found` is present — mirroring how `LoanerJob` already gates on missing data. Purely successful/new-paint runs no longer create an issue.
- Split the job test into two cases: problems → issue created, clean run → no issue.

## Notes

Reuse keys off the last `GithubIssueLog` row and the issue's current state on GitHub. If a previous issue was manually closed, the next problematic run intentionally opens a fresh one so a resolved-and-reclosed issue isn't silently reopened.

## Verification

- `bin/rails test test/jobs/loaders/paints_import_job_test.rb` — 2 tests, green
- `standardrb` clean on changed files

🤖

### DIFF
--- a/app/jobs/loaders/loaner_job.rb
+++ b/app/jobs/loaders/loaner_job.rb
@@ -6,7 +6,8 @@ module Loaders
       # Serialize per task so a clean run's resolve can't close an issue a concurrent
       # problematic run just opened. Both recompute their results while holding the lock.
       ApplicationRecord.with_advisory_lock("loaders:loaner_sync") do
-        missing_loaners, missing_models = ::Rsi::LoanerLoader.new.run
+        result = ::Rsi::LoanerLoader.new.run
+        missing_loaners, missing_models = result
 
         model_ids = ModelLoaner.pluck(:model_id).uniq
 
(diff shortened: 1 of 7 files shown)


mortik:
Addressed the P1 from Greptile in 13a90564c. The digest fast-path was returning *before* the GitHub state check, so a closed/deleted issue whose problem recurred with an unchanged body was never re-reported. Now the short-circuit only applies when the logged issue is still open, so a persistent problem re-surfaces after its issue is closed. Added `GithubIssueCreator` test coverage for the open/closed/deleted/unchanged branches. 🤖

PR #4316, opened 2026-08-11
feat(ships): fill item_prices from a daily UEX sync

mortik:
## Why

`item_prices` is empty everywhere — 0 rows locally and 0 availability across all 243 production models. The old `PrefillItemPrices` migration sourced from `CommodityPrice`/`ShopCommodity`, which were tableless and have since been deleted (the migration itself went in `ca764abcf`). So the Base card's "Sold at?" and "Rental at?" have been rendering nothing for months.

This fills the table from a daily [uexcorp.space](https://uexcorp.space) sync.

## What

- `Uex::Client` — four bulk GETs against `api.uexcorp.uk/2.0`. No auth needed, but they 403 without a `User-Agent`, so one is always sent. Endpoint configured via `config/app/uex.yml` the same way as the RSI one.
- `Uex::VehicleMatcher` — our slugs are already in UEX's format (`orig-100i`), which covers 105 of the 179 priced vehicles; name and `name_full` cover another 66. The remaining 8 get an explicit `MAPPINGS` hash, same approach as `Rsi::LoanerLoader`.
- `Uex::PriceSyncer` — upserts on `(item, price_type, location, time_range)` so re-runs update rather than duplicate, and drops locations UEX no longer lists. Scoped to `Model` item types, so hand-entered rows for components/paints/modules are untouched.
- `Loaders::UexPricesJob` on the `loaders` queue, daily at 05:00 UTC — clear of the existing daily band and well away from the evening loader block.
- A "Powered by UEX" link under the availability lists, shown only when there is availability to credit.

## Two decisions worth a look

**UEX purchase price maps to our `sell`, not `buy`.** Ours is shop-perspective — `ItemPrice#sell?` means the shop sells it, which is what `Model#sold_at` selects and what the UI labels "Sold at?". UEX's is player-perspective. Getting this backwards would silently empty `soldAt` while filling an unused `boughtAt`, so there's a test pinning it.

**Rentals are stored as `1-day`, and this is inferred.** UEX carries one rental row per (vehicle, terminal) with no period field at all. Rent sits in a tight ~2.5% band of the purchase price across the whole set, which matches the in-game 1-day rate:

| Vehicle | Rent | Buy | % |
| --- | ---: | ---: | ---: |
| M50 | 37,485 | 1,499,400 | 2.50% |
| 300i | 34,398 | 1,375,920 | 2.50% |
| 600i Explorer | 680,793 | 27,231,800 | 2.50% |
| Avenger Titan | 27,165 | 1,290,370 | 2.11% |

Nothing in the data can confirm it. **If it turns out to be the 30-day rate, every rental figure is off by ~30×** — worth an in-game check at a rental terminal before this goes live. The fix if so is one constant (`Uex::PriceSyncer::RENTAL_TIME_RANGE`) plus a re-run; no migration.

## Verification

Ran the syncer against the live API in development:

```
created=632 updated=0 removed=0 unmatched=0
288 sell, 344 rental, 0 buy — 179 models with availability
```

A second run reported `created=0 updated=0 removed=0`, so it's idempotent. All 8 mappings resolve. `GET /v1/models/aegs-avenger-titan` comes back with an empty `boughtAt`, two `soldAt` locations and three `rentalAt` rows at `timeRange: "1-day"`.

Backend: 30 new tests, full suite green apart from `Short::ModelCompareShareTest`, which fails identically on a stashed clean baseline (pre-existing parallel-run flake). Frontend: 82 passed, `vue-tsc` clean.

## Notes

- No N+1: the models controller already preloads `:item_prices` on every endpoint that renders a model, and `sold_at`/`rental_at` filter the loaded association in memory.
- Unmatched vehicles open a GitHub issue so the mapping stays maintained as UEX adds ships. The issue body deliberately carries *only* the unmatched list — `GithubIssueCreator` dedupes on a digest of the body, and prices move daily, so including the sync counts would open a fresh issue every single run. Counts go on `import.output` instead.
- Adding `Imports::UexPricesImport` also needs an `ImportTypeEnum` entry and a jbuilder partial, or `ImportTest`'s "every broadcasting import type renders its jbuilder partial" fails.
- UEX's API terms impose no attribution requirement; the credit is a courtesy.
- Out of scope: the availability modal still shows only `location`, not price or rental period, even though both are now in the payload. Also components/paints/modules — UEX has item prices too, but ships first.

Plan and findings at `docs/exec-plans/uex-price-sync.md`.

🤖

### DIFF
--- a/app/api_components/shared/v1/schemas/enums/import_type_enum.rb
+++ b/app/api_components/shared/v1/schemas/enums/import_type_enum.rb
@@ -11,6 +11,7 @@ module Shared
             Imports::ModelImport Imports::ModelsImport Imports::ScData::AllImport
             Imports::ScData::ModelsImport Imports::ScData::ModelImport Imports::HangarSync
             Imports::HangarImport Imports::ModulesImport Imports::PaintsImport
+            Imports::UexPricesImport
           ].freeze
 
           schema({
--- a/app/frontend/frontend/components/Models/BaseMetrics/index.spec.ts
+++ b/app/frontend/frontend/components/Models/BaseMetrics/index.spec.ts
@@ -1,0 +1,59 @@
+import { mount } from "@vue/test-utils";
+import { describe, expect, it, vi } from "vitest";
+import type { Model, ItemPrice } from "@/services/fyApi";
+import Component from "./index.vue";
+
+vi.mock("@/shared/composables/useI18n", () => ({
+  useI18n: () => ({
+    t: (key: string) => key,
+    toNumber: (value: unknown) => String(value),
+    toDollar: (value: unknown) => String(value),
+    toUEC: (value: unknown) => String(value),
+  }),
+}));
+
+const itemPrice = (attributes: Partial<ItemPrice>) =>
+  ({
+    id: "price-1",
+    location: "Astro Armada - Area 18",
+    ...attributes,
+  }) as ItemPrice;
+
+const model = (availability: Partial<Model["availability"]>) =>
+  ({
+    availability: { boughtAt: [], soldAt: [], rentalAt: [], ...availability },
+    classificationLabel: "Multi-Role",
+    metrics: {},
+  }) as unknown as Model;
+
+const mountWith = (availability: Partial<Model["availability"]>) =>
+  mount(Component, {
+    props: { model: model(availability) },
+    global: { directives: { tooltip: () => {} } },
+  });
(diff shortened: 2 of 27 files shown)


mortik:
Superseded by #4317 — same commits, branch renamed `docs/uex-price-sync` → `feat/uex-price-sync`. 🤖

PR #4317, opened 2026-08-11
feat(ships): fill item_prices from a daily UEX sync

mortik:
## Why

`item_prices` is empty everywhere — 0 rows locally and 0 availability across all 243 production models. The old `PrefillItemPrices` migration sourced from `CommodityPrice`/`ShopCommodity`, which were tableless and have since been deleted (the migration itself went in `ca764abcf`). So the Base card's "Sold at?" and "Rental at?" have been rendering nothing for months.

This fills the table from a daily [uexcorp.space](https://uexcorp.space) sync.

## What

- `Uex::Client` — four bulk GETs against `api.uexcorp.uk/2.0`. No auth needed, but they 403 without a `User-Agent`, so one is always sent. Endpoint configured via `config/app/uex.yml` the same way as the RSI one.
- `Uex::VehicleMatcher` — our slugs are already in UEX's format (`orig-100i`), which covers 105 of the 179 priced vehicles; name and `name_full` cover another 66. The remaining 8 get an explicit `MAPPINGS` hash, same approach as `Rsi::LoanerLoader`.
- `Uex::PriceSyncer` — upserts on `(item, price_type, location, time_range)` so re-runs update rather than duplicate, and drops locations UEX no longer lists. Scoped to `Model` item types, so hand-entered rows for components/paints/modules are untouched.
- `Loaders::UexPricesJob` on the `loaders` queue, daily at 05:00 UTC — clear of the existing daily band and well away from the evening loader block.
- A "Powered by UEX" link under the availability lists, shown only when there is availability to credit.

## Two decisions worth a look

**UEX purchase price maps to our `sell`, not `buy`.** Ours is shop-perspective — `ItemPrice#sell?` means the shop sells it, which is what `Model#sold_at` selects and what the UI labels "Sold at?". UEX's is player-perspective. Getting this backwards would silently empty `soldAt` while filling an unused `boughtAt`, so there's a test pinning it.

**Rentals are stored as `1-day`, confirmed in game.** UEX carries one rental row per (vehicle, terminal) with no period field at all. An in-game rental board for the Avenger Titan prices all four durations:

| 1 day | 3 day | 7 days | 30 days |
| ---: | ---: | ---: | ---: |
| 27,165 | 71,308 | 142,616 | 509,344 |

UEX reports 27,165, so `RENTAL_TIME_RANGE = "1-day"` is correct. Worth noting the durations are discounted rather than linear (1× / 2.63× / 5.25× / 18.75×), so this was worth pinning down rather than assuming. UEX carries only the 1-day rate; `ItemPrice#time_range` already enumerates all four, so a future source could add the rest without a migration.

## Verification

Ran the syncer against the live API in development:

```
created=632 updated=0 removed=0 unmatched=0
288 sell, 344 rental, 0 buy — 179 models with availability
```

A second run reported `created=0 updated=0 removed=0`, so it's idempotent. All 8 mappings resolve. `GET /v1/models/aegs-avenger-titan` comes back with an empty `boughtAt`, two `soldAt` locations and three `rentalAt` rows at `timeRange: "1-day"`.

Backend: 30 new tests, full suite green apart from `Short::ModelCompareShareTest`, which fails identically on a stashed clean baseline (pre-existing parallel-run flake). Frontend: 82 passed, `vue-tsc` clean.

## Notes

- No N+1: the models controller already preloads `:item_prices` on every endpoint that renders a model, and `sold_at`/`rental_at` filter the loaded association in memory.
- Unmatched vehicles open a GitHub issue so the mapping stays maintained as UEX adds ships. The issue body deliberately carries *only* the unmatched list — `GithubIssueCreator` dedupes on a digest of the body, and prices move daily, so including the sync counts would open a fresh issue every single run. Counts go on `import.output` instead.
- Adding `Imports::UexPricesImport` also needs an `ImportTypeEnum` entry and a jbuilder partial, or `ImportTest`'s "every broadcasting import type renders its jbuilder partial" fails.
- UEX's API terms impose no attribution requirement; the credit is a courtesy.
- Buy figures match the in-game board too, with one 4 aUEC drift: UEX reports 1,290,370 at New Deal Lorville against 1,290,366 in game. Expected for a community-maintained feed with per-row verification dates.
- Out of scope: the availability modal still shows only `location`, not price or rental period, even though both are now in the payload. Also components/paints/modules — UEX has item prices too, but ships first.

Plan and findings at `docs/exec-plans/uex-price-sync.md`.

🤖

### DIFF
--- a/app/api_components/shared/v1/schemas/enums/import_type_enum.rb
+++ b/app/api_components/shared/v1/schemas/enums/import_type_enum.rb
@@ -11,6 +11,7 @@ module Shared
             Imports::ModelImport Imports::ModelsImport Imports::ScData::AllImport
             Imports::ScData::ModelsImport Imports::ScData::ModelImport Imports::HangarSync
             Imports::HangarImport Imports::ModulesImport Imports::PaintsImport
+            Imports::UexPricesImport
           ].freeze
 
           schema({
--- a/app/frontend/frontend/components/Models/BaseMetrics/index.spec.ts
+++ b/app/frontend/frontend/components/Models/BaseMetrics/index.spec.ts
@@ -1,0 +1,59 @@
+import { mount } from "@vue/test-utils";
+import { describe, expect, it, vi } from "vitest";
+import type { Model, ItemPrice } from "@/services/fyApi";
+import Component from "./index.vue";
+
+vi.mock("@/shared/composables/useI18n", () => ({
+  useI18n: () => ({
+    t: (key: string) => key,
+    toNumber: (value: unknown) => String(value),
+    toDollar: (value: unknown) => String(value),
+    toUEC: (value: unknown) => String(value),
+  }),
+}));
+
+const itemPrice = (attributes: Partial<ItemPrice>) =>
+  ({
+    id: "price-1",
+    location: "Astro Armada - Area 18",
+    ...attributes,
+  }) as ItemPrice;
+
+const model = (availability: Partial<Model["availability"]>) =>
+  ({
+    availability: { boughtAt: [], soldAt: [], rentalAt: [], ...availability },
+    classificationLabel: "Multi-Role",
+    metrics: {},
+  }) as unknown as Model;
+
+const mountWith = (availability: Partial<Model["availability"]>) =>
+  mount(Component, {
+    props: { model: model(availability) },
+    global: { directives: { tooltip: () => {} } },
+  });
+
+describe("ModelBaseMetrics", () => {
+  it("credits UEX when a sale location is shown", () => {
+    const link = mountWith({ soldAt: [itemPrice({})] }).get(
+      'a[href="https://uexcorp.space"]',
+    );
+
+    expect(link.text()).toContain("model.poweredByUex");
+  });
+
+  it("credits UEX when only a rental location is shown", () => {
+    const wrapper = mountWith({
+      rentalAt: [itemPrice({ id: "price-2", timeRange: "1-day" })],
+    });
+
+    expect(wrapper.find('a[href="https://uexcorp.space"]').exists()).toBe(true);
+  });
+
+  it("omits the credit when the ship has no availability", () => {
+    const wrapper = mountWith({});
+
+    expect(wrapper.find('a[href="https://uexcorp.space"]').exists()).toBe(
+      false,
+    );
+  });
+});
--- a/app/frontend/frontend/components/Models/BaseMetrics/index.vue
+++ b/app/frontend/frontend/components/Models/BaseMetrics/index.vue
@@ -22,6 +22,10 @@ const props = withDefaults(defineProps<Props>(), {
 const soldAt = computed(() => props.model.availability.soldAt);
 const rentalAt = computed(() => props.model.availability.rentalAt);
 
+const hasAvailability = computed(
+  () => !!soldAt.value?.length || !!rentalAt.value?.length,
+);
+
 const displayLength = computed(() => {
   if (props.extended && props.model.metrics.extendedLength) {
     return props.model.metrics.extendedLength;
@@ -185,6 +189,13 @@ const displayHeight = computed(() => {
                     {{ modelPrice.location }}
                   </li>
                 </ul>
+              </div>
+            </div>
+            <div v-if="hasAvailability" class="col-12">
+              <div class="metrics-attribution">
+                <a href="https://uexcorp.space" target="_blank" rel="noopener">
+                  {{ t("model.poweredByUex") }}
+                </a>
               </div>
             </div>
           </div>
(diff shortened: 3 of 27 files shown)


mortik:
Good catch on the P1 — valid, and the blast radius was wider than described. Fixed in 550c165bc.

There were four routes to a full wipe, not just a partial one. `/terminals/` or `/vehicles/` returning `status: "ok"` with `data: []` filters every price row out, so `desired` comes back empty and all 632 rows get destroyed. A `data: null` took the same path, because my `Array(payload["data"])` quietly turned it into `[]`. And if UEX ever renamed the `vehicle_buy`/`vehicle_rent` terminal types, the post-filter set would empty out with the raw feed still looking healthy.

Three guards:

1. **Client rejects a non-array `data`** instead of coercing it — a null `data` is malformed, not an empty snapshot.
2. **Syncer refuses an empty feed**, checked *after* the vehicle-terminal filter so a type rename is caught too. None of the four feeds is ever legitimately empty, so this raises and the import is marked failed.
3. **Deletions are capped.** If a run would drop more than `MAX_REMOVAL_RATIO` (0.5) of the prices we hold, it keeps them and reports to AppSignal. Upserts still apply, so fresh prices land and only the destructive half is held back; the next whole snapshot reconciles. 0.5 leaves plenty of headroom for real churn — the largest single vehicle terminal is well under a fifth of all rows.

Verified against the live API: a normal run still reports `created=0 updated=0 removed=0 skipped_removals=0` with all 632 rows intact, and feeding it an empty terminals list now raises instead of wiping.

8 new tests cover it — one per empty feed, the terminal-type rename, a truncated snapshot that must not delete, a minority removal that still must, and the null `data` case. 🤖

mortik:
Right again, and this one exposed that my first fix was the wrong shape. Fixed in f1e32c29e.

The aggregate ratio can't distinguish a real closure from a truncation, and the numbers make that fatal rather than theoretical: on live data Astro Armada is **69 of 632 rows, 10.9%**. Any threshold loose enough to allow genuine churn also allows that terminal's prices to be deleted when it simply doesn't report. A percentage was never going to work here.

So the deletion decision is now made **per terminal**, which is a real signal rather than a proxy:

- A terminal that **priced something** in this snapshot is authoritative for its own inventory — a row of ours it didn't list really is gone, so it's deleted.
- A terminal that **priced nothing** is ambiguous, and the terminals feed settles it. Gone from there too → the shop closed, rows deleted. Still listed → the feed came back short, rows kept and reported to AppSignal with the locations named.

That also fixes a flaw in my own first attempt that you didn't flag: a purely defensive guard would have leaked stale rows forever for a terminal that genuinely closed. The terminals-feed cross-check handles that.

`MAX_REMOVAL_RATIO` stays, demoted to a backstop for the one case the per-terminal rule can't judge — a terminal that reports, but reports short.

Verified against the live API by making Astro Armada go silent:

```
Astro Armada holds 69 of 632 rows (10.9% — under the 50% cap)
terminal silent:  removed=0  skipped_removals=69   → 632 rows, all 69 kept
terminal closed:  removed=69 skipped_removals=0    → 563 rows, correctly dropped
restore:          created=69                       → back to 632
```

Two new tests cover both branches. 37 in the UEX suite now, full backend suite green. 🤖

mortik:
Valid — that residual was real, and I'd consciously left it. Fixed in 42a371216.

My rule asked "did this terminal report **at all**", which is binary, so a terminal that reported a *slice* of its stock still counted as authoritative and its omissions were deleted. The better question is whether it reported at anything like its **usual volume**: a shop discontinuing a ship or two still lists the rest, while a truncated feed shows up as a terminal listing a fraction of what we hold for it — or nothing.

Nicely, that subsumes both earlier mechanisms. A silent terminal is just a 100% drop, so the global `MAX_REMOVAL_RATIO` and the reported-locations set are both gone. Three mechanisms became one rule:

```ruby
def deletable?(location, listed:, held:, live:)
  return true if live.exclude?(location)          # shop closed

  listed.fetch(location, 0) >= held.fetch(location, 0) * MIN_TERMINAL_RETENTION
end
```

Verified against the live API on Astro Armada (69 rows):

```
reports 28/69 (40%):  removed=0   skipped_removals=41   → all 69 kept
reports 52/69 (75%):  removed=17  skipped_removals=0    → plausible churn, applied
drops 1 ship:         removed=1                          → applied
terminal closed:      removed=69                         → applied
```

**One residual I'm accepting deliberately**, flagged here rather than left silent: a terminal truncated to *above* the retention floor still has its omissions deleted. No content-based rule can separate that from a shop genuinely discontinuing that share of its stock — the 75% row above is indistinguishable from reality either way. The rows return on the next good sync. Closing it fully means deleting only after N consecutive absences, which needs a `last_seen_at` column on the shared `item_prices` table; I don't think that's worth a migration for a self-healing 24-hour window, but happy to add it if the maintainer disagrees.

39 tests in the UEX suite, full backend suite green. 🤖

PR #4338, opened 2026-08-12
feat(ui): redesign the Btn component

mortik:
Rebuilds `shared/components/base/Btn` — new visuals, a coherent prop API, and nesting that the button owns itself instead of being restyled from the outside.

Plan and rationale: [`docs/exec-plans/btn-redesign.md`](docs/exec-plans/btn-redesign.md)

## What it looks like

Variant, tone, size, state, group and dropdown matrix

Toolbar and pagination — one surface across the group, dimmed (not lightened) disabled arrows, and the label segment matching its members:

Toolbar and pagination

Sizes line up with the form controls, so a button sits flush next to an input:

Search input and button aligned

## Why

A census over all 468 call sites drove the API:

| Finding | Evidence |
| --- | --- |
| The default size was wrong | `size="small"` at 296 of 468 sites; the old default passed explicitly **twice** |
| `inline` only undid the component's own margin | 112 sites passing a prop whose sole job was cancelling `margin-bottom: 20px` |
| `variant` conflated chrome with intent | `default \| transparent \| link \| danger` — a quiet destructive button was inexpressible |
| Focus was invisible | `outline: none` with no `:focus-visible` replacement |

## Changes

**Visuals** are taken from the metrics-card language (`Models/metricsCard.scss`) rather than designed fresh: `rgba($gray-light, .5)` edge, `$primary` hover, 150ms transitions. The end-caps keep the signature but are retokenised to `#4a4f54` with a **proportional inset** — `max(10px, 18%)` — because the old fixed 14px left only 8px of cap on a 36px icon button.

**API**: `variant` (solid|ghost|bare) × `tone` (neutral|danger) × `size` (sm|md|lg). Heights match `FormInput` (43px, 55px). Removed `inline`, `flush`, `mobileBlock`, `alignStart`, `textInline`. `exact` was dead — vue-router 5 has no such prop. Btn ships no margins; spacing belongs to the container.

**Nesting** is driven by injected context, so `BtnGroup` and `BtnDropdown` no longer reach into Btn's internals — **118 lines of `> :deep(.panel-btn)` overrides with `!important` are gone**. `BtnGroup` follows `metrics-card__hero`: the container owns border and radius, and an inner track clips member corners so no member needs to know its position (which is what broke a `BtnDropdown` trigger — first *and* last child of its own wrapper).

**Accessibility**: `:focus-visible` outline; the label is retained while loading with an `sr-only` status rather than being replaced; a disabled `href` renders a real disabled `<button>` instead of a still-clickable `<a>`; transitions respect `prefers-reduced-motion`.

**Un-leaked** the `panel-btn` overrides from 17 stylesheets. Most were margin arithmetic their `gap` containers already handle, so they're deleted rather than renamed. Two consumers assumed the old inverted hover (light surface, dark content) and are corrected. Deletes two dead files: the embed's near-duplicate of the button styles, and `panel-btn-toggle.scss`, which nothing imported.

## Notable implementation detail

Component styles are plain `<style scoped>`, **not** `lang="scss"`. Verified empirically: under sass, `@apply`/`@reference` pass through untouched and reach the minifier as unknown at-rules, silently dropping the styles. Compiled as plain CSS, each theme token inlines with a literal fallback — `var(--color-edge,#7a828880)` — which is what lets these styles work inside the embed bundle, where `tailwind.css` never loads.

## Testing

`vue-tsc` · `tsc` · `eslint` · `prettier` · `vitest` 82/82 · `vite build` all pass.

New `test/playwright/e2e/Buttons.spec.ts` — 13 assertions, one per property that regressed during the rebuild. Computed-style assertions rather than pixel snapshots: each failure was an invariant violation, and an assertion names the broken invariant where an image diff only says something moved.

⚠️ **The e2e suite was not run.** It doesn't boot the app in a worktree — Rails resolves the asset host to `localhost:3000` while the test Vite server binds `127.0.0.1:3137`, so assets 404 and the SPA never mounts. The pre-existing `Home.spec.ts` fails all 5 of its tests identically, so this is worktree provisioning rather than anything on this branch — but it does mean **`pnpm test:e2e:run` still needs a green run somewhere the suite boots.**

## Review notes

- The 118 sites that previously omitted `size` shrink 48px → 43px. Intentional, and the one change here that isn't behaviour-preserving.
- Embed rendering is worth a browser check: the duplicated `embed/partials/panel-btn.scss` is deleted on the evidence that Vite injects the button CSS via the lazily-imported embed route.
- `$primary` is now the hover signal on every button while also being the primary metrics-tile rail — worth confirming a hovered button and a highlighted stat don't read as the same emphasis.

🤖


mortik:
@greptile-apps review

PR #4339, opened 2026-08-12
feat(visual-tests): add metrics, lists, forms, typography and states pages

mortik:
## What

The visual-test section covered panels, buttons, tables, notifications, the sync modal and the support hint. Everything introduced since — the metrics cards, the inline editable list and the shared form, typography and state primitives — had no page to check it on. This adds five pages to close that gap.

| Page | Covers |
| --- | --- |
| `/visual-tests/metrics/` | `MetricsCard` frame (default + slim, `head` slot, 1/3/6-tile bodies) and its slotted content primitives, `MetricsList`, and the model metrics components (Base/Crew/Speed/Panel/Cargo/Combat/Defense/Hull) fed with live data |
| `/visual-tests/lists/` | `ListGroup`, `InlineEditableList`, `FilteredList` |
| `/visual-tests/forms/` | Every `FormInput` type and variation, `FormInputGroup`, `FormTextarea`, `FormCheckbox`, `FormToggle`, `FormDatePicker`, `FormFileInput`, `RadioList`, `FilterGroup`, `Slider`, `Toggle`, `FormActions` |
| `/visual-tests/typography/` | `Heading` sizes/levels/alignments/modifiers, `Text`, `Pill`, `Spacer`, `Box`, `Grid` |
| `/visual-tests/states/` | `Empty`, `Loader`, `SmallLoader`, `ProgressBar`, `Paginator`, and the three error pages |

A few deliberate choices:

- **The editable list is wired, not mocked.** Create, edit, destroy, selection and the expanded slot all mutate local state, so the states are reachable by clicking rather than by prop-toggling.
- **`FilteredList` gets a synthesised `AsyncStatus`**, with buttons to flip loading / refetching / empty / error.
- **The metrics page uses live API data** for the model components, on `rsi-constellation-andromeda` — an in-game combat ship, so the combat, survivability and hull cards all have something to render. A concept ship has no game-file hardpoints and they all self-hide. There is also a no-data section that asserts exactly that self-hiding.
- **Error states are a child component** (`forms/ErrorStates.vue`) with its own `useForm`, validated on mount, because two `useForm` calls in one setup would clobber each other's field context.

The nav prefixes are renumbered along the way — six entries were all sitting on `"02"`.

## Two bugs the pages caught

Both fixed in [`7bdfceb`](https://github.com/fleetyards/fleetyards/commit/7bdfceb66), separate from the pages themselves:

- **`MetricsCard`'s default variant never set a layout on `__head`** — only `--slim` did — so anything passed to the `head` slot stacked *under* the title instead of beside it. The flex row moves onto the base rule and slim keeps just its padding, border and background. No consumer uses the `head` slot yet, which is why this had never shown up.
- **`MetricsList` used `align-items: flex-center`**, which is not a valid CSS value, so label and value were never vertically centred.

Worth flagging separately: `MetricsList` has no callers anywhere in the app.

## Verification

- `eslint`, `vue-tsc`, `tsc` and `prettier` all clean.
- All five pages driven in a browser: no Vue errors, no missing i18n keys. Remaining console output is pre-existing dev noise (CSP, vite HMR socket, service worker, 401 on `/users/me` while logged out).
- Create / edit / expand / select clicked through end to end on the editable list.
- All 10 metrics cards on the real ship page verified unchanged after the `__head` fix.

Translations added for all 7 locales across `nav`, `headlines` and `title`.

Routes stay dev-only — they are already gated behind `NODE_ENV !== "production"`.

🤖

### DIFF
--- a/app/frontend/frontend/components/Models/MetricsCard/index.vue
+++ b/app/frontend/frontend/components/Models/MetricsCard/index.vue
@@ -117,9 +117,6 @@ withDefaults(defineProps<Props>(), {
 }
 
 .metrics-card--slim .metrics-card__head {
-  display: flex;
-  align-items: center;
-  gap: 12px;
   padding: 14px 16px;
   border-bottom: 1px solid rgba($gray-light, 0.28);
   background: rgba(#000, 0.12);
--- a/app/frontend/frontend/components/Navigation/VisualTestsNav/index.vue
+++ b/app/frontend/frontend/components/Navigation/VisualTestsNav/index.vue
@@ -31,28 +31,58 @@ const { t } = useI18n();
       prefix="02"
     />
     <NavItem
+      :to="{ name: 'visual-tests-typography' }"
+      :label="t('nav.visualTests.typography')"
+      icon="fadt fa-text-size"
+      prefix="03"
+    />
+    <NavItem
+      :to="{ name: 'visual-tests-forms' }"
+      :label="t('nav.visualTests.forms')"
+      icon="fadt fa-input-text"
+      prefix="04"
+    />
+    <NavItem
       :to="{ name: 'visual-tests-tables' }"
       :label="t('nav.visualTests.tables')"
       icon="fadt fa-table"
-      prefix="02"
+      prefix="05"
+    />
+    <NavItem
+      :to="{ name: 'visual-tests-lists' }"
+      :label="t('nav.visualTests.lists')"
+      icon="fadt fa-list-ul"
+      prefix="06"
+    />
+    <NavItem
+      :to="{ name: 'visual-tests-metrics' }"
+      :label="t('nav.visualTests.metrics')"
+      icon="fadt fa-gauge-high"
+      prefix="07"
+    />
+    <NavItem
+      :to="{ name: 'visual-tests-states' }"
+      :label="t('nav.visualTests.states')"
+      icon="fadt fa-spinner"
+      prefix="08"
     />
     <NavItem
       :to="{ name: 'visual-tests-notifications' }"
       :label="t('nav.visualTests.notifications')"
       icon="fadt fa-bell"
-      prefix="02"
+      prefix="09"
     />
     <NavItem
       :to="{ name: 'visual-tests-sync-modal' }"
       :label="t('nav.visualTests.syncModal')"
       icon="fadt fa-arrows-rotate"
-      prefix="02"
+      prefix="10"
     />
     <NavItem
       :to="{ name: 'visual-tests-support-hint' }"
       :label="t('nav.visualTests.supportHint')"
       icon="fadt fa-heart"
-      prefix="02"
+      prefix="11"
     />
   </div>
 </template>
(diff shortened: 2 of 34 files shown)


mortik:
Thanks — three of the four are real and are fixed in cd700e6. The fourth I'm declining, with evidence below.

### 1. Destroy leaves ghost selections — valid, fixed

Reproduced before fixing: select 3 rows, destroy one, list drops to 15 rows while the toolbar still reads `3 Selected`, and the bulk Copy action receives the dead id.

I did **not** take the suggested `resetSelected()`, because that discards the *other* selections too — destroying one row shouldn't deselect the rest. Pruning just the destroyed id instead:

```ts
editableList.value.selected = editableList.value.selected.filter(
  (id) => id !== item.id,
);
```

Worth recording *why* this has to live in the consumer rather than in `InlineEditableList`: selection is deliberately designed to outlive `items` changing, so select-all can span pages —

```ts
// select-all unions with what is already selected (keeps off-page ids)
internalSelected.value = [...internalSelected.value, ...props.items.map((i) => i.id)].filter(uniqArray)
// deselect-all only strips ids on the current page
internalSelected.value = internalSelected.value.filter((id) => !currentIds.includes(id))
```

A blanket prune-on-items-change inside the component would therefore regress paginated select-all. Only the record owner knows a row is gone for good versus merely off-page.

After: 3 selected → destroy one → `2 Selected`, 15 rows, other two selections intact.

### 2. Reset preserves stale list state — valid, fixed

Verified before: open a create row, an edit row and an expanded row, hit Reset → all three stayed open against the restored records. Now closes edit and create, clears selection and collapses the expanded row. After: all false, 16 rows.

### 3. New styles bypass SCSS convention — declining

I don't think this one holds. The convention in `.cursor/rules/frontend/styling.mdc` is:

> - Write custom SCSS instead of relying on **Bootstrap** utility classes.
> - **Align naming and patterns with Tailwind CSS conventions** to ease a future migration to Tailwind.

`flex items-center justify-between` *is* the Tailwind-aligned naming the rule asks for — the thing being discouraged is Bootstrap utilities (`d-flex`, `mb-3`). It's also the established idiom here: 17 occurrences of that exact string across the app, including `admin/pages/models/[id]/edit/docks.vue`, which is the page this fixture is modelled on.

On `.visual-tests-lists__expanded` — the rule says lowercase and dashes, which this is. BEM `__` is used throughout the codebase (`.metrics-card__head`, `.inline-editable-list__toolbar`, `.base-input__wrapper`), so it matches its neighbours.

Happy to reconsider if the intent is that *no* utility classes should be used, but that would be a codebase-wide change rather than something to start in this PR.

### 4. Submit timer outlives page — valid, fixed

Timers are tracked and cleared in `onBeforeUnmount`, on both the fake submit and the toggle round-trip in `forms.vue`, plus the same pattern in `states.vue`'s fixed-loader demo. Verified: submit, navigate away inside the 1.5s window, no stale `Form submitted.` on the destination route.

---

Separately, and out of scope here: `admin/pages/models/[id]/edit/paints.vue` and `.../images.vue` combine `selectable`, `@destroy` and `#selected-actions`, so they have the same ghost-selection issue in production. Worth a follow-up.

🤖

Issue #3788, opened 2026-04-25
Fleet Logistics: Inventory Management & Request System


mortik:
## Summary

Add a logistics system to fleets that enables inventory management and a request workflow for resources and ship outfitting.

### Core Features

- **Fleet Inventories** — Named stockpiles managed by logistics officers. Items can link to existing game data (Models, Components, ModelModules) or use freeform names for commodities not yet in the database.
- **Logistics Requests** — Members can request supplies, ship outfitting help, or general resources. Requests flow through an approval workflow (draft → submitted → approved → in_progress → fulfilled). Officers with logistics privileges are fast-tracked; regular members need approval.
- **Outfitting Requests** — A member can request help outfitting a specific multicrew ship for fleet operations, linking the request to their fleet vehicle so logistics can see cargo capacity and hardpoints.
- **Request Comments** — Lightweight comment thread on each request for coordination between requester and logistics officers.
- **Notifications** — Request lifecycle notifications (submitted, approved, rejected, fulfilled, new comment).

### New Models

| Model | Purpose |
|-------|---------|
| `FleetInventory` | Named stockpile with optional manager assignment |
| `FleetInventoryItem` | Line item in an inventory (polymorphic or freeform) |
| `FleetLogisticsRequest` | Request with AASM state machine and priority levels |
| `FleetLogisticsRequestItem` | Line item in a request with requested/fulfilled tracking |
| `FleetLogisticsComment` | Comment on a logistics request |

### New Privileges

- `fleet:inventories:*` (read, create, update, delete, manage)
- `fleet:logistics:*` (read, create, update, approve, fulfill, delete, manage)

### Request Sources

| Source | Behavior |
|--------|----------|
| Regular member | Creates request → needs approval from logistics/leadership |
| Logistics officer | Creates request → auto-approved |
| Fleet leadership | Creates request → auto-approved |

### Exec Plan

See `docs/exec-plans/fleet-logistics.md` for the full implementation plan (16 phases).

PR #4340, opened 2026-08-12
feat(hangar): personal hangar inventories

mortik:
Brings the fleet logistics inventory system to the personal hangar, behind a `hangar_inventories` flag.

## Why

Fleet logistics (#3788) gave fleets named inventories with an immutable deposit/withdrawal ledger and stock derived by aggregation. Individual players have the same need — tracking cargo, components, ammo and consumables sitting in their own hangars and stations — without having to belong to a fleet.

## What

Multiple named inventories per user (name, description, location, image), each holding the same ledger the fleet feature uses. Per-inventory and hangar-wide stock views, category/quality filters, and CSV import.

| | Fleet | Hangar |
| --- | --- | --- |
| Owner | `fleet_id` | `user_id` |
| Visibility | `members_only` / `officers_only` | private to the owner |
| Manager | `managed_by` | — (owner is the manager) |
| Ledger entry attribution | `member_id`, `added_by` | — |
| Notification on new entry | logistics officers + manager | — |
| Index routes | `/logistics` + `/logistics/inventories` | one merged `/hangar/inventories` |

Reached from an **Inventories** button beside Wishlist and Stats on the hangar page, shown only when the flag is enabled.

## Sharing, not duplicating

Rather than copy the fleet implementation, the reusable half was extracted first and both owners now sit on it:

**Backend** — `InventoryStock` and `InventoryLedgerEntry` concerns hold the stock aggregation, the withdrawal-exceeds-stock validation, the enums and the sorting/ransack config. Each model wires its own foreign key through a class macro:

```ruby
class HangarInventoryItem < ApplicationRecord
  include InventoryLedgerEntry

  inventory_association :hangar_inventory
end
```

`FleetInventoryItemCsvImporter` became `InventoryItemCsvImporter`, with owner-specific entry attributes supplied by the inventory.

**Frontend** — the filter form, inventory panel and the stock/log tables moved to `components/Logistics/` with flags for the owner-specific columns; the two fleet pages now use them and got shorter as a result. New `useInventoryOptions` and `useInventoryStockList` composables replace the option lists and client-side stock filtering both pages had duplicated. The reusable translation keys moved from `fleets.logistics.*` to a top-level `logistics.*`.

## Bug found along the way

Running the app surfaced a defect the tests missed — and the **existing fleet endpoints had it too**:

`ValidationError` looks up `validation_error.<code>`, and none of the inventory codes had translations. Development raises on missing translations, so a rejected withdrawal returned **500 instead of the 400 the API promises**:

```
I18n::MissingTranslationData: Translation missing: en.validation_error.hangar_inventory_items.create
  app/models/validation_error.rb:8:in 'ValidationError#initialize'
```

Fixed for both fleet and hangar inventory endpoints, across all seven locales (`b9dc86c`).

## Verification

125 inventory tests pass — 62 new hangar integration specs, 10 model specs, and the fleet suite unchanged, which is what guards the concern extraction. `vue-tsc`, `tsc`, ESLint, Prettier and Vitest are clean.

Driven against a live server with a real session:

| Check | Result |
| --- | --- |
| Create inventory | 201, slug generated |
| Deposit 10 SCU → withdraw 4 | stock `10.0` → `6.0` |
| Withdraw 100 | 400, `insufficient_stock`, "exceeds current stock (6.0)" |
| CSV import (2 valid, 1 blank name) | `{"imported":2,"errors":[{"row":4,...}]}` |
| Hangar-wide stock across 2 inventories | aggregated per inventory |
| Second user reads/writes the first user's inventory | 404 both |
| Flag off | 403, and absent from `/features` |

**Not verified: the UI in a browser.** Vite dev binds to `fleetyards.test`, which doesn't resolve in my environment, and no Chromium was installed — both fixes touch the host machine, so I left them alone. The pages compile and their server-rendered shells return 200, but nobody has watched them render. Worth a look before merging.

## Unrelated, but you'll hit it

`bin/dev` recurses infinitely in a shell where `./bin` is on `PATH` relatively: `bin/op` strips its own directory by absolute path (`script_dir`), never matches the relative entry, and `exec`s itself forever. Not touched here.

## Follow-ups

- Public sharing of hangar inventories on `/hangar/:username`
- Transferring items between a hangar inventory and a fleet inventory
- The fleet plan's iteration-2 request/approval workflow

🤖


mortik:
@greptile-apps review

mortik:
@greptile-apps review

mortik:
@greptile-apps review

mortik:
@greptile-apps review

mortik:
Reworked the storage layer before this ships: `hangar_inventories` / `hangar_inventory_items` are now `inventories` / `inventory_items` with a polymorphic `holder` (`User` today, `Fleet` once those rows migrate, `Vehicle`-held cargo after that).

The reason to do it here rather than later is timing — these migrations have never run in production, so reshaping them costs a migration rewrite instead of a data migration plus `versions.item_type` and `active_storage_attachments.record_type` rewrites.

Nothing about the API moved: same routes, same payloads, same OpenAPI components, no frontend changes. The integration tests carry over with only factory-argument edits (`user:` → `holder:`), which is the evidence that this is behaviour-neutral. Two jbuilders needed a one-word fix where they called the association by its old name.

Policies keep their context names (`HangarInventoryPolicy`) and are now named explicitly at every `authorize!` call — `Inventory` says nothing about who holds it, so implicit lookup resolving to one shared policy is precisely what should not happen once fleets join the table.

Follow-ups: ship inventories in #4367, fleet inventories migrating onto the same table whenever the next fleet-logistics feature comes along. Plan for both in `docs/exec-plans/vehicle-inventories.md`.

🤖

mortik:
@greptile-apps review

Bypassing the file limit — most of the diff is generated schema and translations. The parts worth scrutiny are narrow:

- `app/controllers/api/v1/hangar_all_inventory_stock_controller.rb` and `hangar_all_inventory_items_controller.rb` — `where(inventories: {holder: current_resource_owner})` is now the only thing separating one user's stock from another's, since inventories no longer live in a user-specific table.
- `app/policies/hangar_inventory_policy.rb` / `hangar_inventory_item_policy.rb` — `Inventory` carries a polymorphic holder, so these are named explicitly at each `authorize!` rather than resolved from the record class.
- `app/models/concerns/inventory_ledger_entry.rb` — the withdrawal stock check locks the inventory row; worth confirming it still holds now that both fleet and personal entries reach it through different associations.

🤖

mortik:
@greptile-apps review

mortik:
@greptile-apps review

mortik:
@greptile-apps review

mortik:
@greptile-apps review

mortik:
@greptile-apps review

PR #4385, opened 2026-08-15
feat(commodities): add Commodity reference table from game files

mortik:
Adds a `Commodity` reference table sourced from the Star Citizen game files, as the first of several reference tables backing the fleet inventory feature.

## Why

`FleetInventoryItem` stores a free-text `name` and has an unused polymorphic `item` association. This gives that association something to point at for `category: commodity`, so members can pick a real commodity instead of typing one.

## What's here

- **`ScData::Parser::CommoditiesParser`** → `data/sc_data/parsed/live/commodities/*.json` (168 commodities)
- **`ScData::Loader::CommoditiesLoader`** → upserts into the new `commodities` table
- **`Commodity` model + migration**, registered in `BaseParser.all` / `BaseLoader.all` so the existing `Loaders::ScData::AllJob` picks both up

## Notes on the parse

Commodities are declared across **two** record trees, not one. `entities/commodities/` looks like the list but holds only 131 — the per-crate-size records under `entities/scitem/carryables/` cover 165, including every ore and harvestable. The union is the complete set.

Type resolution needed a precedence rule because the game data disagrees with itself:

- The alloys (Steel, Atlasium, Dymantium) are tagged both `alloy` and `metal` across crate records
- Contraband variants of legal goods claim `vice`
- ~38 commodities declare no `displayType` at all — mostly harvestables and tractor-beam-only crates

So: the `displayType` on the `entities/commodities/` record wins (it's what the commodity kiosk reads), then most-common declared type, then canonical folder, then record-path tokens. Result is **0 untyped and 0 unresolved conflicts**, verified in the loader test.

Two game-data bugs are handled: Wuotan Seed's display name points at its description key, and `items_commodities_type_processedGoods` is the generic label on unmarked crates rather than a commodity (dropped — hence 168, not 169).

## Loaded data

168 rows — mineral 30, metal 23, natural 22, vice 16, consumer_goods 16, processed_goods 14, manmade 12, gas 12, food 6, agricultural_supply 4, alloy 4, nonmetals 3, medical_supply 3, scrap 2, waste 1. 155 have descriptions, 131 have a canonical game ref.

## Design decisions

- **Keyed on the game localization key** (`sc_key`), nullable so a UEX-only commodity can exist without a game record.
- **`uex_id` / `uex_slug` are nullable and unpopulated.** Prices, supply/demand and terminal inventories are datacore-only and not in the extracted XML — those come from UEX, and this is the join column for when they do.

## Not in this PR

- The API endpoint, the picker in `InventoryItemModal`, and an `item_type` allowlist — all in progress on this branch next.
- Storing the localized type label (agreed, coming next).
- Icons. `Commodity#icon` holds a path that currently resolves to nothing: the game-data exporter only extracts `*.xml` and `*.ini`, so the referenced SVGs were never pulled. Filed as fleetyards/sc-tools#1.
- Translated commodity names. All 11 locales are available in the dump, but no sc_data is translated anywhere in the app today, so this stays English-only.

## Testing

`bin/rails test test/models test/loaders` → 182 tests, 0 failures. 10 of those are new (model + loader), covering type resolution for refined metals, ores, harvestables and the alloy conflict, plus loader idempotency.

🤖

### DIFF
--- a/app/lib/sc_data/loader/base_loader.rb
+++ b/app/lib/sc_data/loader/base_loader.rb
@@ -8,6 +8,7 @@ module ScData
         ::ScData::Loader::ItemsLoader.new.all
         ::ScData::Loader::ModelsLoader.new.all
         ::ScData::Loader::ModelModulesLoader.new.all
+        ::ScData::Loader::CommoditiesLoader.new.all
       end
 
       def initialize
--- a/app/lib/sc_data/loader/commodities_loader.rb
+++ b/app/lib/sc_data/loader/commodities_loader.rb
@@ -1,0 +1,35 @@
+module ScData
+  module Loader
+    class CommoditiesLoader < ::ScData::Loader::BaseLoader
+      def all
+        load_items("commodities").each do |commodity_data|
+          one(commodity_data)
+        end
+      end
+
+      def one(commodity_data)
+        return if commodity_data["name"].blank?
+
+        commodity = Commodity.find_by(sc_key: commodity_data["sc_key"])
+        commodity ||= Commodity.find_by(name: commodity_data["name"], sc_key: nil)
+        commodity ||= Commodity.new(sc_key: commodity_data["sc_key"])
+
+        commodity.update!(update_params(commodity_data))
+
+        commodity
+      end
+
+      private def update_params(commodity_data)
+        {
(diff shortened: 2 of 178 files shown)


mortik:
@greptile-apps review

PR #4388, opened 2026-08-15
fix(setup): give each worktree its own test database

mortik:
Worktrees get their own dev database but share one test database, so two suites running at once fight over it.

## The bug

`config/database.yml` already reads the suffix for both environments:

```erb
development:
  database: <%= ENV["DB_NAME"] || "fleetyards_dev#{ENV['WORKTREE_SUFFIX']}" %>
test:
  database: <%= ENV["DB_NAME"] || "fleetyards_test#{ENV['WORKTREE_SUFFIX']}#{ENV['TEST_ENV_NUMBER']}" %>
```

But `bin/setup` writes `WORKTREE_SUFFIX` only to `.env.local`, and **dotenv deliberately skips `.env.local` in the test environment** — that's by design upstream, so local overrides can't quietly change what the suite tests.

The result: dev is correctly isolated (`fleetyards_dev_wt_commodities`) while every worktree's suite runs against one shared `fleetyards_test`. Whichever run tries to purge it second dies:

```
ActiveRecord::StatementInvalid: PG::ObjectInUse:
  ERROR: database "fleetyards_test" is being accessed by other users
  DETAIL: There is 1 other session using the database.
```

Worth noting the port hid half of this: `DB_PORT` never reached the test env either, and it only worked because `8271` happens to be the hardcoded default in `database.yml`.

## The fix

`bin/setup` also writes `.env.test.local`, which dotenv *does* load in test:

```
# -- worktree-managed --
DB_PORT=8271
WORKTREE_SUFFIX=_wt_commodities
```

Only the two variables that decide which database the suite talks to are repeated. The rest of the managed block is deliberately left out — the Redis and Postgres ports are the same shared containers for every worktree, and `DOMAIN`, `PORT` and `COOKIE_PREFIX` would change what the tests themselves assert.

The marker-preserving logic is now shared by both writers rather than duplicated, so user content above `# -- worktree-managed --` is still kept in either file.

## Verification

- `RAILS_ENV=test` now resolves to `fleetyards_test_wt_commodities`
- Ran the suite while another worktree's suite was mid-run — previously an immediate `PG::ObjectInUse`, now it just runs
- `bin/rails test` → 1787 tests, 2 failures + 2 errors, all four pre-existing on `main` and unrelated (2 WebMock blocking `vite-test/entrypoints/email.scss`, 2 order-dependent short-link redirects). Those two errors persist on an isolated database, which confirms they were never contention.

## Notes

- **CI is unaffected** — it has no `.env.test.local` (the whole `.env*` glob is gitignored bar the template) and keeps using `fleetyards_test`.
- **Existing worktrees need `bin/setup` re-run** to get their own test database. The old shared `fleetyards_test` is then orphaned and can be dropped whenever.
- Sidekiq's Redis database is still shared across worktrees in test (`SIDEKIQ_REDIS_DB` is per-worktree but only in `.env.local`). I left that alone rather than widen the change — it hasn't caused a collision, and changing Redis wiring under the suite deserves its own look.

🤖

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -130,20 +130,23 @@ def write_superset_ports(port, vite_port)
   File.write(File.join(superset_dir, "ports.json"), JSON.pretty_generate(config) + "\n")
 end
 
-def write_env_local(port, db_port, redis_port, vite_port, db_suffix, sidekiq_redis_db, cookie_prefix: nil)
-  env_local = File.join(APP_ROOT, ".env.local")
+def write_managed_env(filename, managed_body)
+  path = File.join(APP_ROOT, filename)
 
   # Preserve user content above the marker
   user_content = ""
-  if File.exist?(env_local)
-    lines = File.readlines(env_local)
+  if File.exist?(path)
+    lines = File.readlines(path)
     marker_idx = lines.index { |l| l.strip == WORKTREE_MARKER }
     user_content = lines[0...marker_idx].join if marker_idx
     user_content = lines.join unless marker_idx
   end
 
-  managed = <<~ENV
-    #{WORKTREE_MARKER}
+  File.write(path, user_content + "#{WORKTREE_MARKER}\n" + managed_body)
+end
+
+def write_env_local(port, db_port, redis_port, vite_port, db_suffix, sidekiq_redis_db, cookie_prefix: nil)
+  write_managed_env(".env.local", <<~ENV)
     PORT=#{port}
     DB_PORT=#{db_port}
     REDIS_URL=redis://localhost:#{redis_port}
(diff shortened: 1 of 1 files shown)


mortik:
### On the remaining 4/5

All three inline threads are resolved. The score is held down by the summary alone, and its last two objections can't both be satisfied:

- **Round 3 (P1):** a marker glued to a user line must be *recognised*, so the stale managed block below it gets cleaned up.
- **Round 4 (P1):** a line that merely *ends* with the marker must **not** be treated as the boundary, because everything below it would be discarded.

Those describe the same string: `<user text># -- worktree-managed --`. Nothing in the file distinguishes "the old writer glued this on" from "a developer typed it", so any lookup that satisfies one objection violates the other.

I've resolved it by dropping the recovery attempt rather than picking a side. `237a6da3b` returns the boundary lookup to exact-line matching, which is what `main` does today:

```ruby
marker_idx = lines.index { |l| l.strip == WORKTREE_MARKER }
user_content = marker_idx ? lines[0...marker_idx].join : lines.join
```

The reasoning:

- The glued shape can only be produced by the bug this branch fixes (`user_content + marker` with no separator), and only when a hand-edited `.env.local` had its trailing newline stripped. The newline guard means it can't happen again.
- Its consequence was cosmetic, not destructive: a stale block above the marker. dotenv is last-wins, so the live values were always the fresh ones. It also stabilises — after one run the file ends with a real marker line, which the next run finds and truncates at. It does not accumulate a block per run.
- Recovering it, by contrast, required treating a line the writer never wrote as the boundary, which is a path to deleting real developer config. That's a worse trade for a cosmetic gain, and it was scope creep on a PR about per-worktree test databases.

A file already in the glued state is now left alone: nothing deleted, one fresh managed block appended, stable on every subsequent run.

Verified against a harness that evals `write_managed_env` out of `bin/setup` — 10 cases, all passing: fresh file, rerun idempotency, user content above the marker, trailing whitespace after the marker, missing trailing newline, marker text mid-line, a user line ending in the marker text, and a pre-glued file across two runs. `standardrb` clean.

🤖

PR #4391, opened 2026-08-16
feat(equipment): personal gear reference table from game files

mortik:
Personal gear as an inventory reference table: weapons, melee, attachments and magazines, parsed from the game files and pickable in an inventory entry.

## One table, because the game already has one

`AttachDef Type` splits every record into `WeaponPersonal` (426) and `WeaponAttachment` (172) — and **magazines sit under `WeaponAttachment`**, beside iron sights and barrels. Ammunition isn't a separate catalogue to fold in; in CIG's taxonomy it was never separate. So `weapon`, `equipment` and `ammunition` ledger entries all reference `Equipment`.

Scope is the six personal trees (`fps_weapons`, `melee`, `weapon_modifier`, `magazines`, `throwable`, `mines`) — 623 records. Turrets and hardpoint-mounted guns stay with `Component`.

## Reshaping the dormant table rather than adding one

`Equipment` existed, empty, with `# TODO: appears unused — remove entirely`. Its columns turn out to fit the game files well — `rate_of_fire`, `range`, `weapon_class`, `item_type` are exactly what the records carry — so it's reshaped: game-file identity columns added, the three enums converted to strings, the store-scraped image columns dropped.

Strings rather than enums for the same reason `Commodity#commodity_type` is one: a class CIG adds in a later build should load rather than raise. There's a test for it.

> **Please confirm `Equipment.count.zero?` in production before merging.** The table is empty in dev and the model says it's unused, but the migration replaces columns, so that's worth checking rather than assuming.

## The descriptions are a spec sheet

Most open with a hand-written block:

```
Manufacturer: Behring
Item Type: Assault Rifle
Class: Ballistic

Magazine Size: 40
Rate Of Fire: 550 rpm
Effective Range: 50 m
```

Reading it fills `rate_of_fire`, `range` and `storage` for around 300 records each — up from 17/17/68 — and gives `assault_rifle` where the record key only says `rifle`. The block spans several paragraphs and attachments label it `Type:` rather than `Item Type:`, so it took two passes; descriptions still holding a stray spec line went 134 → 10. What's left is prose.

## Hiding skins without a suffix list

623 records carry only 535 distinct names — twelve are all "Pyro RYT Multi-Tool". Rather than grow a list of `_tint`/`_collector`/`_ai`/`_tow` suffixes every patch, the rule is structural:

> a key that extends another key **and carries the same display name** is a copy of it

The name test is what keeps real items in — a rifle's magazine key also extends the rifle's, but it's named for the magazine, so it survives. There's a test asserting exactly that.

Visible rows sharing a name: **23 → 4**. The last four are CIG genuinely reusing a name across two makers' guns, which is why the picker labels options with their size.

## Icons

`Equipment#icon` records the `displayIcon` path. Worth knowing before anything is built on it: **185 of 382** visible items name one, there are only **28 distinct images** because every variant of a weapon shares its family icon, and **no attachment names one at all** (there's a test pinning that, so it'll tell you if that changes). The assets aren't in the export, so nothing resolves today — this is stored for when they are.

## Two bugs fixed on the way

- **The reshape migration silently dropped two indexes.** It created the `equipment_type` and `item_type` indexes before replacing those columns, and `remove_column` takes a column's index with it. Verified fixed by rebuilding the database from migrations.
- **The equipment list wasn't sorted.** `ransack_alias :name` points `name` at `name_or_slug` so a search matches either, which leaves ransack unable to sort by it. Ordered on the relation instead, with a comment saying why.

## Testing

34 new tests. `bin/rails test` → **1997 tests**, 2 failures + 2 errors, all four pre-existing and unrelated (2 WebMock blocking `vite-test/entrypoints/email.scss`, 2 order-dependent short-link redirects). `pnpm test` 191 passed; `lint:ts`, `lint:fix` and `standardrb` clean.

## Follow-up

Armour and clothing (`characters/human/{armor,clothing,body,head,starwear}`) join this table rather than getting their own — which is why `slot`, `damage_reduction`, `temperature_rating` and the compatibility columns are kept despite nothing filling them yet.

🤖


mortik:
@greptile-apps review

PR #4392, opened 2026-08-16
feat(equipment): armour and clothing

mortik:
Armour and clothing join the same `Equipment` table as the weapons — worn gear rather than a second kind of thing.

> **Base:** `feat/equipment-reference-table` (#4391). Merge that first; this then rebases onto `main` cleanly.

## What lands

**4775 records, 4441 visible**, from `characters/human/{armor,clothing,starwear}` on top of the weapon trees:

| Family | Visible |
|---|---|
| armor | 2133 |
| clothing | 1861 |
| weapon | 239 |
| undersuit | 202 |
| weapon_attachment | 143 |

`AttachDef Type` names the slot for worn gear — `Char_Armor_Helmet`, `Char_Clothing_Feet` — and fills **all twelve** values of the `slot` enum, including the torso's two clothing layers (`Torso_0` → shirt, `Torso_1` → jacket).

## The dormant columns were sketched for this

The armour spec block lines up with what the table already had:

```
Item Type: Heavy Utility
Damage Reduction: 25%
Temp. Rating: -225 / 75 °C
Radiation Protection: 33600 REM
Radiation Scrub Rate: 147.42 REM/s
Carrying Capacity: 8.0 µSCU
Backpacks: All
```

`damage_reduction`, `temperature_rating`, `core_compatibility` and `backpack_compatibility` map straight across — filled for 2168, 2210, 122 and 469 records. Radiation protection, scrub rate and G-force tolerance had no column, so this adds three.

`Temp. Rating` stays a string: it's a range, not a figure.

## Generalising the skin rule

The rule from #4391 only caught keys that *extend* another key, which missed clothing entirely — the seventeen `mym_shirt_01_01_NN` are siblings, not extensions, and all read "Davlos Shirt (Charcoal)". A picker would have listed it seventeen times.

Now: **records sharing a display name and a maker-and-type key prefix are variants of one item**, and only the first by key stays visible. Visible duplicates went **101 → 19**.

Both halves of that test earn their place, and there are assertions for each: without the name test a rifle's magazine would be hidden by the rifle it shares a prefix with; without the prefix test the two makers who each ship a "BR-2 Shotgun" would collapse into one.

## A bug the existing tests caught

Parsing `180K µSCU` meant reading a thousands suffix — and my first attempt was case-insensitive, so `Effective Range: 50 m` became fifty million. The weapon spec test failed immediately. The suffix is now upper-case only and must not be the first letter of a longer unit, which also keeps `147.42 REM/s` from matching.

## Tests

The loader tests now read the parsed files rather than loading 4,700 rows per test — **198s → 43s**, with more assertions than before, and they're a truer description of what's under test (the parser's output, not the loader's writes). Three genuine loader tests remain: it loads, it resolves manufacturers, it's idempotent.

`bin/rails test` → **2001 tests**, 2 failures + 2 errors, all four pre-existing and unrelated. `standardrb` clean, OpenAPI schema regenerated.

## Also here

The public endpoint now exposes the armour stats — `slot`, `damageReduction`, `temperatureRating`, the radiation pair, `gForceTolerance` and both compatibility fields. They're the interesting part of an armour record, and the endpoint was returning none of them.

## Follow-up

Consumables (`entities/scitem/consumables`, `consumabletypesdatabase`) are the last of the reference tables.

🤖


mortik:
@greptile-apps review

PR #4398, opened 2026-08-16
chore(cleanup): prune the components older builds left behind

mortik:
Components accumulate a row per build forever. This prunes them, without touching the ones players still depend on.

## The growth

`Component` is keyed on **(sc_key, version)**, not `sc_key`:

```ruby
component = Component.find_by(sc_key: normalized_key, version: sc_version)
...
component = Component.create!(sc_key: normalized_key, sc_ref: ref, version: sc_version) if component.blank?
```

So an import never updates the existing row — it adds a fresh one per component, and nothing has ever taken the old ones away. `Component.current_version` exists precisely because the table is assumed to be full of history; it just grows without bound.

Worth knowing before merging: `Component.group(:version).count` will tell you how much this is actually costing in production. I can't see that from here.

## Why it can't just delete

Four tables carry a `component_id`, and one of them is user data:

| Reference | |
|---|---|
| `hardpoints.component_id` | game data, rebuilt each import; FK to components |
| `model_hardpoints.component_id` | game data, `dependent: :nullify` |
| `model_hardpoint_loadouts.component_id` | game data, **no FK and no association back from Component** |
| **`vehicle_loadout_hardpoints.component_id`** | **players' saved loadouts, FK to components** |

A saved loadout points at the component it was built with, and that reference is what keeps the loadout meaning what its owner chose. The foreign key would refuse the delete in any case — the good failure mode, but it means a naive cleanup would simply error.

So the job **keeps the current build and whatever players still reference**, and takes the rest. That bounds growth to one build plus a small tail.

`model_hardpoint_loadouts` is cleared explicitly, because nothing else would: it has neither a foreign key nor an association back from `Component`. And rows are destroyed rather than deleted, so a component takes its own sub-hardpoints with it and the hardpoints fitted with it get their reference cleared before the row goes.

## The guard

The job refuses to run before the current build has been imported. Without it every row looks stale and the table is emptied — a scheduled job with that failure mode is worse than no job. There's a test for it.

## Also here

**An index on `components.version`.** There wasn't one — only `manufacturer_id` — so every `current_version` scope was a filtered scan over the growing table. Added concurrently.

Scheduled weekly on Sunday at 03:00, after a week of imports have landed.

## Testing

Six tests, covering what a data-deleting job needs to get right: stale rows go, a component a saved loadout references stays, the model-hardpoint-loadout reference is cleared, sub-hardpoints go with their component, nothing happens before the current build is imported, and nothing happens when there is nothing stale.

`bin/rails test` → 2007 tests, 2 failures + 2 errors, all four pre-existing and unrelated.

🤖

### DIFF
--- a/app/jobs/cleanup/component_versions_job.rb
+++ b/app/jobs/cleanup/component_versions_job.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module Cleanup
+  # Components are keyed on (sc_key, version), so every imported build adds a
+  # fresh row per component and nothing ever took the old ones away. The rows
+  # cannot simply be dropped: a saved vehicle loadout points at the component it
+  # was built with, and that reference is what keeps the loadout meaning what
+  # its owner chose.
+  #
+  # So this keeps the current build and whatever players still reference, and
+  # takes the rest.
+  class ComponentVersionsJob < ::Cleanup::BaseJob
+    Result = Struct.new(:removed, :kept_for_loadouts) do
+      def to_s
+        "removed=#{removed} kept_for_loadouts=#{kept_for_loadouts}"
+      end
+    end
+
+    def perform
+      version = Rails.configuration.sc_data[:version]
+
+      # Before the current build has been imported every row looks stale, and
+      # the table would be emptied. Nothing to clean up until it lands.
+      return if version.blank? || Component.where(version:).none?
+
+      referenced = VehicleLoadoutHardpoint.where.not(component_id: nil).select(:component_id)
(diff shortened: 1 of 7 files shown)


mortik:
Superseded by #4399.

This pruned rows that shouldn't exist in the first place — "keep whatever a loadout still references" was a workaround for loadouts pointing at build-specific rows. #4399 removes the cause instead: one row per component, updated in place, with the spec history kept as PaperTrail versions.

The `components.version` index from here is still worth having and has been folded into #4399, placed after the collapse so it builds over the table that survives rather than all 36k rows.

🤖

PR #4399, opened 2026-08-16
refactor(components): one row per component, history as versions

mortik:
A component is one row now, updated in place, with its spec history kept as PaperTrail versions instead of as a row per build.

## What it replaces

Components were keyed on `(sc_key, version)`, so an import never updated a row — it added another. Production is carrying the result:

| version | rows |
|---|---|
| 4.9.0-live.12344265 *(current)* | 7306 |
| 4.8.0-live.11825000 | 7124 |
| 4.7.1-live.11638371 | 6681 |
| 4.7.1-live.11592622 | 6681 |
| 4.7.0-live.11518367 | 6626 |
| *(none)* | 1663 |
| **total** | **36,081** |

That is also why a saved vehicle loadout could end up pointing at a build-specific row, and why `Component.current_version` had to exist at all.

`Commodity` and `Equipment` already work the way this moves `Component` to: one row per `sc_key`, `version` recording the build it was last seen in.

## History without the growth

```ruby
has_paper_trail on: %i[update], only: %i[name description size grade ...]
```

Scoped the way `Model` already does it, and **`version` is deliberately not tracked** — an import touches every component and moves that column on all of them, so recording it would write a history row per component per run and bury the changes worth reading.

## The migration

Collapses the existing rows in three passes following the loader's own lookup order — `sc_key`, then `sc_ref` for rows with no key, then `name` for rows with neither — so what merges is what the loader would now treat as one component. The passes are scoped not to overlap.

The survivor is the row from the current build, falling back to the newest. Everything pointing at the others — `hardpoints`, `model_hardpoints`, `model_hardpoint_loadouts`, `vehicle_loadout_hardpoints`, and `item_prices` — is repointed at it first. A unique index on `sc_key` then stops the table growing back.

**A component the current build no longer ships keeps its row.** Only duplicates of the same component collapse; nothing is dropped for being old, because a saved loadout may still reference it.

## Two bugs worth knowing about

**PaperTrail silently broke the `version` column.** PaperTrail defines a `version` reader, which shadows this table's `version` **column**: `create!` worked, but every `update!` wrote `NULL`. Left in, `current_version` would have matched nothing after the first import. The associations are renamed to `paper_trail_version` / `paper_trail_versions`.

Worth noting for `Commodity` and `Equipment` — both have a `version` column, so the same trap is waiting there if PaperTrail is ever added.

**The survivor ordering picked the wrong row for versionless components.** `ORDER BY (version = 'current') DESC` yields `NULL` for the 1663 rows carrying no version, and Postgres sorts NULLs *first* under `DESC` — so one of those would have beaten the current build. Now `COALESCE(version = 'current', FALSE)`.

That one only surfaced because of the production counts above; it would not have shown up against clean data.

## Verified

Against a fixture built to production's shape — one component across four builds plus a versionless row, a component the current build dropped, and two keyless rows:

- 7 rows → 3
- the current build's row is the survivor, not the versionless one
- the dropped component keeps its row
- prices and sub-hardpoints repointed, orphaned sub-hardpoints removed

Plus tests that a spec change writes a version, that a version-only bump does not, that `sc_key` is now unique, and that the loader updates a past build's row rather than adding another.

`bin/rails test` → 2005 tests, 2 failures + 2 errors, all four pre-existing and unrelated.

## Before merging

Worth running on staging first: the migration ends by creating a **unique** index on `sc_key`, so if production holds duplicates the three passes don't catch, that fails the deploy rather than the migration.

Supersedes #4398, now closed — it pruned rows that should not exist under this model. Its `components.version` index is folded in here, placed after the collapse so it builds over the table that survives.

🤖

### DIFF
--- a/app/lib/sc_data/loader/base_loader.rb
+++ b/app/lib/sc_data/loader/base_loader.rb
@@ -112,13 +112,15 @@ module ScData
 
           if item["key"].blank? && item["ref"].blank? && parent.is_a?(Model) && item["name"]&.end_with?("_module")
             derived_key = "#{parent.sc_data_identifier}_module"
-            item["key"] = derived_key if Component.exists?(sc_key: derived_key, version: sc_version)
+            item["key"] = derived_key if Component.exists?(sc_key: derived_key)
           end
 
+          # Resolved by key alone: a component is one row now, and `version`
+          # says which build it was last seen in rather than which row to use.
           component = if item["key"].present?
-            Component.find_by(sc_key: item["key"]&.downcase, version: sc_version)
+            Component.find_by(sc_key: item["key"]&.downcase)
           elsif item["ref"].present?
-            Component.find_by(sc_ref: item["ref"], version: sc_version)
+            Component.find_by(sc_ref: item["ref"])
           end
 
           if component.present?
@@ -135,9 +137,9 @@ module ScData
                                  nested = nested_loadout.find { |nl| nl["name"]&.downcase == hp.sc_name }
                                  if nested.present?
                                    sub_component = if nested["key"].present?
-                                     Component.find_by(sc_key: nested["key"]&.downcase, version: sc_version)
+                                     Component.find_by(sc_key: nested["key"]&.downcase)
                                    elsif nested["ref"].present?
-                                     Component.find_by(sc_ref: nested["ref"], version: sc_version)
+                                     Component.find_by(sc_ref: nested["ref"])
                                    end
                                  end
                                end
--- a/app/lib/sc_data/loader/items_loader.rb
+++ b/app/lib/sc_data/loader/items_loader.rb
@@ -62,6 +62,10 @@ module ScData
             end
           end
 
+          # Stamped on every load, not just on create: the row is reused across
+          # builds now, so this is what records the build it was last seen in.
+          update_params[:version] = sc_version
+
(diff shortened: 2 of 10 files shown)


mortik:
Confirmed and fixed in 58156d0.

`Hardpoint has_many :hardpoints, as: :parent, dependent: :destroy` — a turret owns the mount, which owns the gun fitted to it. Raw SQL gets none of that cascade, so deleting only the rows whose `parent_type` was `Component` left the children behind pointing at a parent that no longer existed.

Now a recursive CTE takes the whole subtree:

```sql
WITH RECURSIVE doomed AS (
  SELECT id FROM hardpoints
  WHERE parent_type = 'Component'
    AND parent_id IN (SELECT id FROM component_survivors WHERE id <> survivor_id)
  UNION ALL
  SELECT child.id FROM hardpoints child
  JOIN doomed ON child.parent_type = 'Hardpoint' AND child.parent_id = doomed.id
)
DELETE FROM hardpoints WHERE id IN (SELECT id FROM doomed)
```

Verified against a fixture with three levels hanging off a collapsed component — mount, gun, barrel. All three go, and afterwards nothing points at a missing parent or a missing component. Under the previous query only the mount matched, leaving the other two orphaned, which is exactly what was reported.

🤖

PR #2655, opened 2023-08-20
chore: Migrate to Vue 3

mortik:
## Summary

- Full Vue 3 migration with Composition API (`<script setup>`) across all components
- Doorkeeper OAuth2 integration with JWT tokens, OpenID Connect, scopes, and refresh tokens
- Admin + user self-service OAuth application management (CRUD)
- Admin RSI API status page with request log management
- Feature flags admin management and user self-service
- Replace manual API clients with Orval-generated ones
- Action Policy authorization (replacing CanCanCan)
- Various frontend/backend improvements and fixes

## Resolves

Closes #2708 — Doorkeeper setup is fully included and extended in this branch (JWT, OIDC, scopes, admin/user OAuth app management).

Resolves: #3239, #3240, #3241, #3242, #2568, #3243, #3244

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4406, opened 2026-08-16
data(sc_data): manufacturer logos, commodity icons, and the artwork itself

mortik:
Parser output only — no code. Produced by #4409, which is the twenty-odd hand-written files that generate it.

Split this way because the two together were unreviewable: 265 generated files around 22 written ones, well past the 100-file mark where Greptile stops reading and past the point where a person can see the change either.

## Parsed manufacturers, 121 files

Re-parsed now that the records' `Logo` attribute is read rather than skipped. **112 of the 120 carry one**, stored as the path the record names, downcased to match what the other catalogues already hold:

```diff
   "name": "Talon",
   "short_name": null,
-  "description": null
+  "description": null,
+  "icon": "ui/sharedassets/manufacturerlogos/talon_256.tif"
```

`manufacturers/asad.json` goes. The folder clearing from #4401 caught it on its first real run: the build ships `ASD`, that file said `ASAD`, and it had been sitting in the tree since the Vue 3 migration in #2655.

## The artwork, 144 files, 2.1 MB

104 manufacturer logos and 40 commodity icons, copied from the PNGs and vectors the export ships — byte-identical to what CIG produced.

They are carried here rather than fetched from the bucket at load time because of the ratio: **2.1 MB referenced against 128 MB of asset trees**. Every variant of an item shares its family artwork, so what the app references is a rounding error on what is exported. At that size, committing costs less than S3 credentials in production and a bucket dependency on every load.

🤖


mortik:
@greptile-apps review

PR #4407, opened 2026-08-16
feat(equipment): read how much room a piece of gear takes

mortik:
> Stacked on #4410, which carries the re-parsed equipment tree. This PR is the code only — 12 files. Base retargets to `main` automatically once #4410 merges.

## What

`equipment.volume` has existed since the armour and clothing import and nothing ever wrote to it — every row was `NULL`. This reads it out of the game files, along with the box each piece physically fills.

## Why two numbers

The game states both on every `AttachDef`, and they answer different questions:

```
inventoryOccupancyVolume    84000 microSCU        -> 0.084 SCU
inventoryOccupancyDimensions 0.447 x 1.106 x 0.802 -> 0.203 SCU as a box
```

The volume is what a cargo grid or an inventory charges for the piece. The box is the ceiling once it snaps to a grid and stops sharing space with its neighbours. Together they bracket "how much room is left", which is what this unblocks — a grid's `capacity_scu` is exactly its own bounding volume over 1.25³, so the units line up.

## Coverage

**4,815 of 4,821** rows measured: armour 2,142, clothing 1,859, weapons 405, undersuits 202, attachments 161, hacking tools 27, medical 19.

The six left out — two clothing, four attachments — declare one microSCU inside a 0.15 m cube, which is the pair CIG leaves on a record nobody has measured; creature bodies and ship storage boxes carry the same one. Stored as `nil` rather than as vanishingly small, so a reader can tell "unmeasured" from "tiny".

## Notes

- The column was `decimal(15, 2)`, which rounded every piece of gear to zero. It moves to six places — a helmet is 0.0032 SCU.
- `volume_dimensions` is new (`jsonb`), matching how `Component#inventory_consumption` already keeps its box.
- Both are exposed on the Equipment API component.

## Testing

`bin/rails test test/loaders/sc_data/equipment_loader_test.rb test/integration/api/v1/equipment_test.rb test/models/equipment_test.rb` — 45 tests, 140 assertions, green, against a test database rebuilt from `schema.rb`.

Five new loader tests cover the coverage floor, that a measured piece carries both numbers, that the placeholder stays blank, and that both persist.

🤖

### DIFF
--- a/app/api_components/shared/v1/schemas/equipment.rb
+++ b/app/api_components/shared/v1/schemas/equipment.rb
@@ -26,6 +26,19 @@ module Shared
             range: {type: [:number, :null]},
             storage: {type: [:number, :null]},
 
+            # What the piece costs a cargo grid or an inventory, and the box it
+            # fills once it snaps to one and stops sharing space.
+            volume: {type: [:number, :null]},
+            volumeDimensions: {
+              type: [:object, :null],
+              properties: {
+                x: {type: :number},
+                y: {type: :number},
+                z: {type: :number}
+              },
+              additionalProperties: false
+            },
+
             damageReduction: {type: [:number, :null]},
             temperatureRating: {type: [:string, :null]},
             radiationProtection: {type: [:number, :null]},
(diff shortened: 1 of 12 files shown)


mortik:
@greptile-apps review

PR #4416, opened 2026-08-17
feat(hangar): ship inventories

mortik:
Stacked on #4415 — the parser output lives there, so this PR is code only.

Replaces #4375, which GitHub would not let me retarget (*"cannot change the base branch because the pull request is part of a stack"*). Same branch, same commits, review history is on the old thread.

Hangar inventories answer "what is in this locker". They do not answer the question people actually have: *the cargo is in the Ironclad — where is the rest of my stuff?* A ship's cargo hold is the same bucket with a different holder, so this hangs inventories off vehicles instead of introducing a third subsystem.

Plan: [`docs/exec-plans/vehicle-inventories.md`](docs/exec-plans/vehicle-inventories.md).

## Data model

`inventories` gains `vehicle_id` with three partial indexes. `holder` stays *who owns it*; `vehicle_id` is *where it is*. Keeping them apart is what lets slice 3 express "org cargo aboard a member's Hull C", which a `holder: Vehicle` design cannot say.

One inventory per vehicle, because implicit provisioning only works if "where does this deposit go" has exactly one answer. Two Ironclads therefore produce two inventories both called "Ironclad Inventory" — fine, since nothing resolves by that string any more. The name is stored rather than derived, so an inventory orphaned by a sale keeps a name that still means something, and `on_delete: :nullify` plus a `before_destroy` copies the ship's name into `location` before the link goes.

## Provisioning

The inventory does not exist until something is put in it.

- `GET /vehicles/:id/inventory` on an untouched ship returns `200` with `id: null` and creates nothing — a GET that writes would break caching and would fill a 60-ship hangar with empty buckets.
- The first `POST .../items` creates it, inside the same transaction as the entry, so a rejected deposit leaves no empty inventory behind.
- Two first deposits racing are settled by the unique index on `vehicle_id`; the insert sits in a savepoint so the loser can still read the winner's row.
- `DELETE .../inventory` drops the row and its entries. That is "clear this ship's cargo" and needs no separate concept.

## API

`/vehicles/:id/inventory` with `items`, `stock` and the CSV import nested under it, addressed by vehicle id or serial the way loadouts already are (`vehicles.slug` is nil for most ships). Scoping the lookup to the signed-in user's vehicles is the authorization boundary, so no policy branches on holder type.

Two things worth flagging for review:

- **`HangarInventoryItemsController` and `HangarInventoryStockController` were refactored onto a shared concern** before the vehicle ones were written, so ship inventories cost three thin controllers instead of a second copy of ninety lines. The hangar suite passes **unmodified** — that commit is the proof.
- **Slug lookups in the hangar controllers are now scoped to inventories without a vehicle.** Ship inventories share a slug with every other ship of the same model, so `/hangar/inventories/:slug` would otherwise be ambiguous.

New OpenAPI components are neutral (`Inventory`, `InventoryItem`, `InventoryVehicle`, …) and render from the same jbuilders as the `Hangar*` ones, which cannot drift as a result. The hangar names stay put — renaming them breaks generated clients; `HangarInventory` only *gains* a nullable `vehicle`, which is an addition, not a rename.

## Frontend

A Cargo tab on the vehicle page, wiring the existing Logistics components: stock and log tables, deposit/withdraw, CSV import, per-position edit and delete, clear-cargo. Capacity shows `312 / 400 SCU` from `models.cargo` and turns red past the limit, but **nothing rejects a deposit** — players stash cargo outside the grid, and personal inventory is not cargo.

The two Hangar modals were the only components that knew which endpoints they wrote to; they now take an `InventoryTarget` union, which is what keeps slice 3 from needing a third copy of them. On `/hangar/inventories`, ship inventories carry the ship as their subtitle and link to that ship's Cargo tab.

## Verification

- 66 new integration tests across the ten vehicle endpoints, mirroring the hangar suite: another user's ship → 404, unauthenticated → 401, and the import action inside the `hangar:write` guard — the gap that produced a 500 on the fleet side.
- `assert_no_difference "Inventory.count"` on the GET, on the items index, and on a rejected first deposit.
- A lost-race test that drives the real `RecordNotUnique` path inside a surrounding transaction.
- Full backend suite green apart from four pre-existing failures I confirmed on a clean tree (WebMock blocking a Vite asset fetch during mailer rendering, plus one parallel-order flake that passes in isolation). Frontend: 153 tests, `lint:ts` and `lint:fix` clean.

## Added since the first review

- **A cargo position now has its own page**, `/hangar/:id/cargo/:item`, matching hangar and fleet. Stock and log rows link by name; edit and delete moved off the table row, which also removed an artifact — `Btn` draws end-caps for standalone controls, and against a table row they read as bars floating above and below the icon.
- **A route into it at all.** Nothing linked the vehicle page, so the context menu gained *Manage Cargo* beside *Manage Loadouts*.
- **`meta.feature` now blocks navigation.** It only ever hid nav entries, so every flagged route in the app was reachable by URL and answered 403s. The router resolves the flag in `beforeResolve` and sends those to the 404, after the session checks. The three hangar inventories routes declared no flag at all and now do.
- **`ship_inventories` reaches past the vehicle endpoints.** While it is off, the hangar list, the aggregate stock and the aggregate ledger fall back to `Inventory.hand_made` — otherwise the overview kept listing ship inventories and linking them to a tab that was not there.
- **A ship's own storage counts toward capacity.** `models.personal_inventory` is read from the vehicle entity's container (#4415 carries the parsed side), and capacity is now grid + storage rather than the grid alone.
- **Gear counts toward what is aboard.** `totalVolumeScu` measures per-piece gear through the catalogues, so a hold of armour no longer reads as empty, and the tab shows what is left. A position nothing has measured is left out rather than guessed at, and `unmeasuredCount` says how far off the floor might be.
- **The equipment picker is scoped to its category.** One table holds guns, armour and attachments, so picking *Weapon* offered all 103 item types; it now reaches the 24 that are weapons. *Ammunition* is relabelled *Attachments & Ammunition*, since it maps to the whole `weapon_attachment` type.
- **Feature flags go through the generated enum everywhere**, including route meta and `FeatureGuard` props, with `x-enumNames` added so dashed flags get a dot-accessible key.

## Still not in this PR

Consumables cannot link to a catalogue item. #4403 landed the records after this branch was written; wiring the picker is a follow-up.

Slice 3 — migrating fleet inventories onto this table, which unlocks the combined hangar + ships + fleets search — is next and unblocked by the concern extracted here.

🤖


mortik:
Closing — opened in error. #4375 stays the review thread. 🤖

PR #4431, opened 2026-08-17
feat(admin): commodity and equipment reference tables

mortik:
Adds commodities and personal gear to the admin panel — list, create, edit, destroy, filter sidebar, store image, and the item price tab equipment already had rows for — then carries the same treatment across to the components list so the three catalogues read alike.

### Prices

All three price through the one polymorphic `item_prices` table and each was asked the same three questions, so the association, `bought_at`/`sold_at` and the new cheapest-of-each-direction readers collapse into `ItemPriceConcern`. Commodity gains prices with it — the UEX snapshot has been writing rows against `item_type: "Commodity"` since the sync landed and the model had no association to read them back.

The lists show the cheapest buy and sell price and filter on a range of each. The filter goes through a ransacker on a scalar subquery rather than a join: a join on a polymorphic `has_many` repeats the item once per price row, and the `distinct` that fixes that fights the ordering ransack applies on top. `type: :decimal` is what makes `_gteq` compare numbers instead of the query string's text. Same expression behind the payload field and the filter, so the number you filter on is the number the row shows.

That missing association also explains a gap on the **public** API: commodities were the only one of the three catalogues without `availability`, not by choice but because there was nothing to render from. Model prices from the same UEX feed are already public, so there was no contract reason to withhold these. The public commodity payload now carries `boughtAt`/`soldAt` like its siblings.

One fix that came out of this rather than being asked for: these payloads cache on `["v1", record]`, and the snapshot writes prices without touching the item it prices, so cached `availability` was serving the previous day's numbers — on the public equipment and component endpoints as well as the admin ones. The cache keys now carry a price fingerprint: count plus a microsecond-resolution max `updated_at`, since a re-priced row within the same second went unnoticed at second resolution.

### Filter groups

`commodityType`, `equipmentType`, `itemType`, `weaponClass`, `componentClass` and the component `itemType` become pickers instead of substring matching. The backing actions mostly existed but were absent from the schema, so no generated client could reach them. Declaring them means they authorize through their resource policy like the rest of the resource, where before any signed-in admin could read the lists — a deliberate behaviour change on the four pre-existing actions.

Weapon class options are read off the table rather than from `Equipment::WEAPON_CLASSES`. The column is a free string precisely so a class a patch introduces loads rather than raises, so sourcing the picker from the constant would leave those rows unfilterable until someone edited it. Same rule `equipment_types` and `item_types` already follow, with a test for it.

`Admin::V1::Schemas::Component` is new only because there was none — the admin document referenced the shared definition directly, so the two price fields would otherwise have landed in the public schema. Same name, admin scope, no reference changes.

### The widened item type enum

`ItemPriceItemTypeEnum` gains `Equipment` and `Commodity`, which oasdiff reports as 384 `response-property-enum-value-added` warnings across 33 public endpoints — the enum sits under `availability` on models, modules, paints, components and equipment, so two values reach a lot of paths.

Both values were already being served before this change: public `GET /equipment` renders `availability`, and admin `GET /item-prices` lists `ItemPrice.all`, which has held commodity rows since the sync landed. `response_validation` is disabled, so the mismatch went unreported — the enum was too narrow for data the API already returned, and widening it corrects the contract rather than changing what ships.

Ignored per message in `swagger/v1/oasdiff-ignore.txt`, the way the `cosmetic` and inventory `item/type` additions already are, so a future enum addition still has to be looked at rather than passing silently. It is a long block; a `--severity-levels` file would collapse it to one line if you'd rather change that policy.

### The scroll fix

Worth a look in a browser, since it's CSS-only and has no coverage. The list tables were in the group setting `overscroll-behavior: none` on both axes. The y value blocks propagation to the ancestor **even on an axis the container cannot scroll**, so as soon as a table was wide enough to become a scroll container, hovering it froze the page. Every table gets this once it gains a column; the three lists here just crossed the threshold. They scroll sideways only — nothing caps their height — so `x` is the axis that wants `none`. `.compare-table` stays on both: it caps its own height and genuinely owns the vertical axis.

I left `overflow: auto` on the table wrapper alone. Pinning it to `overflow-y: hidden` would also help, but the actions column renders a `BtnDropdown` on mobile that overflows the wrapper vertically, and `hidden` would clip it unreachable where `auto` at least lets you scroll to it.

### Notes

- The four price labels are translated in all seven locales. The filter-group labels reuse keys this branch already left in English in the non-English locales, so they're still pending a translation pass. Five placeholder keys were pruned where the input became a group and nothing reads them any more.
- Payload weight is worth a thought on the public commodities index: UEX prices a commodity at every terminal that trades it, and the endpoint is index-only with `paginates_per 60`. It follows the components precedent, and the controller preloads `item_prices`, but it is the heaviest of the three.
- 2006 tests pass locally (`standardrb`, `pnpm lint:ts`, `pnpm lint:fix` clean); the two local `users_*` mailer errors reach for a Vite asset over HTTP and pass on CI.

🤖


mortik:
@greptile-apps review

mortik:
The camelCase finding is incorrect — `Middleware::TransformParameters` normalizes it, app-wide, before any controller sees the request.

It is registered in `config/application.rb:68` and deep-transforms **both** `request_parameters` and `query_parameters` from camelCase to snake_case, so `equipmentType` arrives as `equipment_type` and the snake_case `permit` list matches. That is why the same camelCase-schema/snake_case-permit pairing already works for every existing admin resource, including the components controller these two were modelled on.

It was also already covered, if incidentally: `PUT /equipment/:id assigns a manufacturer` sends `body: {manufacturerId: manufacturer.id}` and asserts the manufacturer is assigned. That test passes on this branch, which it could not if multi-word keys were being dropped.

That said, the review pointed at a real gap in the tests, so it earned its keep: the `POST` cases only sent `name`, which reads identically in both conventions, so nothing actually exercised the transform. Added in c832a7bc — one test per resource that posts `equipmentType` / `itemType` / `subType` / `weaponClass` / `manufacturerId` / `scKey` and `commodityType` / `uexId` / `uexCode` / `scKey`, asserting both the response and the persisted record.

No production code changed, since none needed to. 🤖

mortik:
@greptile-apps review

PR #4434, opened 2026-08-17
feat(paints): component icons, and curated artwork onto components

mortik:
Stacked on #4435, which carries the parser change and the parsed tree this reads. **Merge #4435 first** — the loader test here asserts against an icon that only exists in that tree.

## `icon` is a separate attachment

`store_image` on a component is curated — an admin upload, or what the hangar sync brought in — so a load has nowhere to put the export's picture without overwriting somebody's work. `has_one_attached :icon` gives it one: the load owns the icon and only the icon, and can rewrite it on every build.

Paints are the first catalogue where the game ships artwork for a record whose `store_image` is maintained by hand, which is what forced the split. There's a test asserting the loader fills `icon` and leaves `store_image` untouched, and another asserting an item whose manufacturer is a *real* one gets no icon at all.

## `bin/paints` — copying curated artwork the other way

The two sides name a paint differently. The game says `Vanguard Clawed Steel Livery` — base ship, paint, suffix — while a ModelPaint is `Clawed Steel` against a Model whose own name carries the variant, `Vanguard Hoplite`. So the paint name is matched as a suffix and what precedes it has to read as the head of the model.

The ship half is what makes a match unique: `Fortuna` is sold on twenty-six ships, so a paint name alone identifies nothing.

```
bin/paints match-report                    # 645 of 945 matched, the rest listed by name
bin/paints copy-store-images --dry-run     # 643 would copy
bin/paints copy-store-images
```

Two behaviours worth reviewing:

- it never overwrites a component that already has a picture — one that does is curated by definition
- it duplicates the blob rather than attaching the paint's own; sharing one would mean purging a paint takes the component's picture with it, which is exactly what consolidating `model_paints` away would do

## Tests

```
paint_component_matcher_test   7 tests,  7 assertions, 0 failures
items_loader_test              6 tests, 15 assertions, 0 failures   (3 new)
```

The matcher tests cover the variant prefix, preferring the base ship over a variant carrying the same paint, and refusing to match a paint of the right name on the wrong ship.

## Not in this PR

`ModelPaint` has no `component_id`, so nothing is persisted — `bin/paints` matches at call time. A stored link is the natural next step, and the nullable FK doubles as the flag for the paints RSI sells before a build carries them. 🤖


mortik:
@greptile-apps review

PR #3893, opened 2026-05-09
feat(fleets): the events and calendar UI

mortik:
Final PR of the stack. Stacked on #4449 → #4448 → #4447 → #4446 → #4445 → #4443 — **review those first**; this diff assumes all of them.

This PR used to be the whole mission builder (455 files, +43k). It has been split into seven; what remains here is the events and calendar frontend.

### What's left here

- The events surfaces: list, calendar grid (month/week), detail page, tabbed editor, and the panels for sign-ups, teams, ships and slots
- `VehiclePicker` with ship-match warnings, the slot picker modal, unassigned sign-ups, occurrence overrides
- Fleet settings pages for calendar, Discord and notifications, plus the personal equivalents
- Short links for events, the visual-test page, and the exec-plan docs (~1.8k lines of the diff)

### The stack

| PR | Scope |
|---|---|
| #4443 | the `fleet_mission_builder` flag |
| #4445 | missions: migrations, models, API |
| #4446 | fleet events: model, state machine, sign-ups, recurrence |
| #4447 | calendar feeds + ICS |
| #4448 | notifications + the Discord sync |
| #4449 | the mission builder UI + shared form/tab components |
| **this** | the events and calendar UI |

### How the split was verified

- The three files that had to be divided across #4446/#4447/#4448 — `fleet_events_controller.rb`, `fleet_events/_base.jbuilder` and the `FleetEvent` schema component — reassemble **byte-identically** to this branch's originals.
- The generated OpenAPI spec reaches **237 paths at #4448, exactly this branch's set**, with none missing.
- This branch's tree is **unchanged** by the restructure except in one place: the flag is declared as `fleet_mission_builder` from the start, instead of adding `mission_builder` and renaming it afterwards. That pair of data migrations is gone.

### Verification on this branch

- Ruby suites — 2688 tests, 6350 assertions, 0 failures
- `vitest` — 33 files, 235 tests passed
- `vue-tsc` + `tsc` clean; `bin/generate-schema` reproduces the committed schema with zero drift

🤖

PR #4446, opened 2026-08-18
feat(fleets): fleet events

mortik:
Third of the stack splitting #3893. Stacked on #4445 → #4443 — **review those first.**

Schedules a mission onto a fleet's calendar. An event carries its own copy of the mission's teams, ships and slots, so editing the event never touches the template it came from.

### What's here

- `fleet_events` with a publish / lock-signups / start / complete / cancel state machine
- Teams, ships and slots, expandable from a model
- Sign-ups with direct or approval flows, per-slot assignment, and event admins who can manage them
- Recurring events as RRULE-on-event: occurrences are derived rather than materialised, with per-occurrence override rows (`fleet_event_occurrence_states`) for the ones that differ
- `FleetEvents::AutoLockJob` closes sign-ups as the start window approaches (5-minute cron)
- Privileges follow missions: `fleet:events:read` for members, `fleet:events:manage` for anyone already managing the fleet, backfilled onto existing roles

### Why this one is bigger than planned

I'd intended to split events core from sign-ups and recurrence, but they aren't separable: the state machine, the sign-up counters and the occurrence generation all live inside `fleet_event.rb`. Cutting them apart would have meant authoring intermediate versions of that model that never existed on the original branch. Feeds and the Discord push *are* separable — they're wired through subscribers and their own files — so those follow in the next two PRs.

Consequently **`ics` and `sync_to_discord` are deliberately absent** from `fleet_events_controller`, along with the Discord fields on the event payload. PR 4 adds the first back, PR 5 the other two.

### Two things to know

- `20260505200000_add_event_columns_to_fleets` adds `calendar_feed_token` alongside `default_timezone`. The column is unused until PR 4 — I kept the migration whole rather than splitting one migration file across two PRs.
- **`swagger/v1/schema.yaml` is generated — skip it.** The path set is PR 2's 201 plus exactly 28 new event paths, nothing removed; the rest of the diff is openapi-ruby reordering.

### Verification

- `test/integration`, `test/models`, `test/lib`, `test/jobs`, `test/mailers` — 2541 tests, 5940 assertions, 0 failures
- `standardrb` clean on every touched Ruby file, `vue-tsc` + `tsc` clean
- `db/schema.rb` regenerated by migrating a scratch database, not dumped from a dev DB, so it contains only this PR's tables

🤖

### DIFF
--- a/app/api_components/v1/schemas/fleets/events/fleet_event.rb
+++ b/app/api_components/v1/schemas/fleets/events/fleet_event.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+module V1
+  module Schemas
+    module Fleets
+      module Events
+        class FleetEvent
+          include OpenapiRuby::Components::Base
+
+          STATUSES = %w[draft open locked active completed cancelled].freeze
+          VISIBILITIES = %w[members officers fleet].freeze
+          SIGNUP_APPROVALS = %w[direct confirmation_required].freeze
+          VIEWER_EVENT_ROLES = %w[creator admin moderator].freeze
+
+          schema({
+            type: :object,
+            properties: {
+              id: {type: :string, format: :uuid},
+              fleetId: {type: :string, format: :uuid},
+              missionId: {type: :string, format: :uuid},
+              title: {type: :string},
+              slug: {type: :string},
+              description: {type: :string},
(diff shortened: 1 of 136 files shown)


mortik:
@greptile-apps review

PR #4449, opened 2026-08-18
feat(fleets): the mission builder UI

mortik:
Sixth of the stack splitting #3893. Stacked on #4448 → #4447 → #4446 → #4445 → #4443. First of the two frontend PRs.

Missions get their own surfaces: a list, a detail page showing the teams and the events spawned from the template, and a tabbed editor for the basics, the description and cover, and the teams with their ships and slots.

### Two commits

1. **`feat(models): filter focus options by classification`** — the ship picker offers focuses for a chosen classification, so the filter endpoint narrows its options the same way the model list does. I'd initially set this aside as an unrelated rider on the branch; `vue-tsc` proved otherwise — `ShipModal` calls `useModelFocusFilters` with a `classification` param, which doesn't exist without it.
2. **`feat(fleets): the mission builder UI`** — the surfaces themselves, plus the shared pieces they introduced: `FormTabs`, `TabNavView`, the `FormInput`/`FormDateTime` additions, `BreadCrumbs`' prev/next stepper, `Timezones`, and the mission category colours in `tailwind.css`. Those shared components are here rather than in the next PR because the events surfaces build on the same ones.

### One judgement call to flag

**Translations for the events surfaces are in this PR too**, unused until the next one. Splitting the i18n trees key by key risked silently missing a dynamic lookup (`t(\`labels.…\${code}\`)`), and no test would have caught it. Shipping the extra keys early is inert; a missing key is a visible bug.

Everything is behind the `fleet_mission_builder` flag from #4443, gated at the nav item and the route wrapper.

### Verification

- `vitest` — 31 files, 223 tests passed
- `vue-tsc` + `tsc` clean, `prettier --check app/frontend` clean, `standardrb` clean
- Ruby suites — 2688 tests, 6350 assertions, 0 failures

🤖


mortik:
@greptile-apps review

PR #4451, opened 2026-08-18
feat(sc-data): serve commodity icons as SVG instead of rasterizing them

mortik:
## What

The game export ships its commodity icons as SVG. We were rasterizing them at parse time — 512px PNGs — because ActiveStorage refuses to serve a vector inline. This carries them through as vectors instead: **256 KB against 960 KB**, and sharp at any density.

Two commits, because the second one is only safe because of the first.

### `fix(uploads)`: reject SVG on every attachment a user can write

Serving vectors inline is a **global** ActiveStorage setting, not a per-attachment one. An SVG can carry a script, and a blob served inline runs it on the storage origin — which shares a registrable domain, and therefore the Devise cookie (`domain: ".#{app.domain}"`), with the app. Today nothing validates content types on *any* user upload, so flipping that setting without this would turn every avatar and fleet logo into a place to park stored XSS.

`NoVectorImageValidator` covers all eight user-writable attachments: `User#avatar`, `Fleet#logo`/`#background_image`, `Inventory#image`, `FleetInventory#image`, the ledger entry image on `InventoryItem` + `FleetInventoryItem`, and `Imports::HangarImport#import` (not a picture, but a user upload whose blob URL the admin UI links to).

Two details worth a look in review:

- It checks `attachment_changes`, i.e. only what *this save* attaches. Validating the attachment itself would leave any record that already carries an SVG permanently unsaveable.
- `OmniauthCallbacksController` skips a vector avatar *before* downloading it. Without that guard a provider serving an SVG fails `user.save` and drops the user into the "username taken" redirect — losing them a sign-in over a picture they never picked.

Admin-writable attachments (model/paint/module store images, `Manufacturer#logo`, `Image#file`) are deliberately left open: the sc_data loaders must keep attaching vectors, and a load can't tell its own attachment from an admin's.

### `feat(sc-data)`: serve the export's icons as vectors

- `BaseParser#save_icon` is a plain copy now; `rasterize`, `RASTER_WIDTH` and `librsvg2-bin` on the CI runner are gone.
- `content_types_to_serve_as_binary -= svg` stops the blob being stamped `application/octet-stream`; `content_types_allowed_inline += svg` stops it being handed over as a download. Both are needed — verified with `blob.service_metadata`, which now comes back `{content_type: "image/svg+xml"}` with no forced disposition.
- `variable_content_types` is deliberately untouched: a vector has no sizes to resize to, and vips marks `svgload` untrusted anyway (`compare_image.rb:164` already works around that).
- `api/v1/shared/_file.jbuilder` gains a vector branch. A vector isn't `representable?`, so `small/medium/large/xlargeUrl` would all be missing — and `smallUrl` is what every panel draws, so the icons would drop back to the placeholder they used to fall through to. All four now point at the blob. No schema change, no frontend change.
- `data/sc_data/parsed/live/icons/` re-parsed from the same `4.9.0-live.12344265` dump: 40 PNGs out, 40 SVGs in. Manufacturer logos are byte-identical.

## Testing

Full suite green — 2444 tests, 0 failures. (Two `Short::ModelCompareShareTest` errors locally are a missing Vite test manifest, `autoBuild: false` in `config/vite.json`; they pass in CI, which builds the frontend.) `standardrb` clean.

New coverage: the eight user-writable attachments are driven off one table, so an attachment added to a user-facing model without the validation fails a test rather than opening a quiet hole; a pre-existing SVG stays saveable; clearing an attachment isn't mistaken for an attach; the omniauth skip; and a commodities API test pinning all four sized URLs to the blob URL — the exact absence that sent these icons to the placeholder last time.

## Noticed but not included

Re-running the parser over the same dump the committed tree was built from also deletes **66 `parsed/live/items/*.json` and 23 `parsed/live/models/*.json`** — drift between the committed tree and the current `blacklisted_item_key?` filters, nothing to do with icons. I restored those folders so this PR is the icon swap and nothing else. Worth doing deliberately: `retire_absent` would set `version: nil` on those 23 models on the next load, dropping them out of `current_version`.

🤖


mortik:
@greptile-apps review

mortik:
On the review's one blocking concern — that enabling inline SVG delivery also exposes user-controlled SVG blobs accepted before the validator existed. The mechanism is real, but it doesn't hold for this deployment, on two independent counts.

**The stored objects can't go inline.** `config/storage.yml` declares the cloud service `public: true`. In `ActiveStorage::Service#url` that takes the `public_url(key)` branch, which — unlike `private_url` — passes no `response_content_type` and no `response_content_disposition`. The browser gets whatever headers the S3 object itself carries, and those are stamped once, at upload, from `Blob#service_metadata`. A blob uploaded before this PR was stamped `application/octet-stream` with `Content-Disposition: attachment`, and a config change neither re-uploads nor re-stamps it. So a pre-existing user SVG stays inert in production regardless of this setting.

Where it *would* bite is the local Disk service, which computes those headers per request from the current config. That's development only.

**There are none anyway.** Run against production:

```ruby
ActiveStorage::Blob.joins(:attachments)
  .where("active_storage_blobs.content_type LIKE ?", "image/svg%")
  .where(active_storage_attachments: {
    record_type: %w[User Fleet Inventory FleetInventory InventoryItem FleetInventoryItem Import]
  })
  .group("active_storage_attachments.record_type", "active_storage_attachments.name")
  .count
# => {}
```

(`Import` rather than `Imports::HangarImport` — polymorphic attachments store the STI base class.)

Worth being explicit about the alternative the finding implies, since it's the one thing here that would be a genuine regression: making the validator check the attachment rather than `attachment_changes` would leave any record that already carries an SVG permanently unsaveable — the user could never edit their profile again, and couldn't fix it by replacing the avatar either, because the save that replaced it would be the save that fails. Ignoring unchanged attachments is the deliberate choice, not an oversight.

🤖

mortik:
@greptile-apps review

mortik:
@greptile-apps review

PR #4468, opened 2026-08-22
chore(sc_data): move the parsed tree out of git into object storage

mortik:
Moves `data/sc_data/parsed` out of git and into the bucket the raw trees already live in, and has the loader jobs fetch the environment they are about to read.

## Why

The parsed tree was **21,605 tracked files / 119 MB**, delivered to production by being baked into the image via the Dockerfile's `COPY . .`.

- **`.git` is 1.9 GB.** Every clone pays for it.
- **A build bump was unreviewable.** The last one — `8e0ab44fe` — was 824 files and 91,705 insertions, carrying the whole tree alongside the two-line version pointer. Nobody reads that diff; they read the loader change buried next to it.
- **The web and API containers carried data they never read.** All four read sites are inside `ScData::Loader::*`; nothing in the request path touches the tree. It is build input for a background job, not application data.

A second environment (PTU) would have doubled all of that, at a higher landing cadence — so this is a prerequisite for that work.

## What changes

A build bump becomes the two-line `config/app/sc_data.yml` diff, with the payload travelling separately:

```
bin/scdata sync              # raw  <- bucket   (unchanged)
bin/scdata parse live        # raw  -> parsed   (unchanged)
bin/scdata push live         # parsed -> bucket (new)
git commit                   # 2 lines, not 1,116 files
                             # then: CheckJob -> AllJob -> pull -> load
```

Worktrees symlink the tree from the main checkout via `bin/setup`, exactly as `data/sc_data/raw` already does.

## Commits

Ordered so the tree only leaves git *after* the tooling and fetch step exist — no commit mid-stack is a state where production could not get its data.

| | |
| --- | --- |
| `b8bcfcedc` | drop the stale `live-preview` (stamped 4.0, Dec 2024) and `probe` fossil trees — 6,529 files |
| `a3cf6b8e6` | `ScData::ParsedStore` + `bin/scdata push`/`pull` |
| `007efba72` | the three loader jobs pull before loading |
| `e6a20ccfb` | stop tracking the tree — 15,076 files |
| `e5df0ab1c` | fall back to the object-storage config already deployed |

## Configuration

Production needs **one value** added to the encrypted credentials:

```yaml
sc_data_s3:
  bucket: fltyrd-sc
```

Endpoint and keys fall back to the existing `s3_*` credentials used by `config/storage.yml`. The bucket differs (`fltyrd-live-storage` is ActiveStorage's, and is `public: true` — game data should not go there), but the account and endpoint do not, and object-storage keys here are project-scoped, so the deployed pair reaches `fltyrd-sc` too. Each `sc_data_s3` key remains an override for a separate account later.

Local development needs nothing: `.env.tpl` supplies all four via `op://` refs, and run bare the CLI falls back to reading 1Password directly the way `sync` does.

## Design notes

**Why a Ruby mirror instead of `s3cmd sync`.** `s3cmd` is Python and the runtime stage (`ruby:*-slim`) has none — measured on Debian slim, adding it costs **19 packages and ~35 MB** to make the same command available to the loader jobs. `aws-sdk-s3` is already in the Gemfile and already loaded in production by ActiveStorage.

**Why it mirrors rather than copies.** A file the source stopped carrying is removed from the target. `ModelsLoader#update_in_game_flag` reads a ship's in-game-ness off file existence, so a leftover from an earlier build would keep a retired ship flying. An empty remote listing raises rather than mirroring the emptiness onto disk, which would otherwise retire the whole catalogue in one run.

**Why unchanged files are skipped.** Local MD5 against the object ETag — it takes a re-push from 3m10s to 6s, and keeps the daily `CheckJob` from re-downloading 119 MB when nothing moved.

**Why the version is verified.** Pointer and payload now live apart and can disagree. `AllJob` checks the tree's `version.json` against the version it was handed, because loading the wrong tree would stamp every row with a version it does not describe.

**The fetch is skipped when the bucket is not configured**, so a developer who just ran `bin/scdata parse` can still load offline with no credentials.

**The commodity matcher test** globbed the parsed tree to check its mappings resolve. With the tree absent on CI that assertion would have passed *vacuously*, silently dropping the guard — so it reads a committed keys fixture instead, rewritten on every `bin/scdata parse` so it cannot drift.

## Verification

The live tree is already pushed — **15,076 objects** at `parsed/live/`; the raw version folders were untouched, since the mirror only addresses keys under its own prefix.

```
push (initial)     Uploaded: 15076, removed: 0        3m10s
push (repeat)      Uploaded: 0, unchanged: 15076      6s      <- idempotent, bare invocation
pull (no-op)       Downloaded: 0, unchanged: 15076    4s
pull (from empty)  Downloaded: 15076, removed: 0      3m04s   <- full restore, tree deleted first
```

That last one was unplanned: rebasing across `e6a20ccfb` deletes the local tree, because git still considers those files tracked at that point. `bin/scdata pull live` restored it from scratch — which incidentally exercised the exact path the loaders will take.

**Worth knowing after merge:** anyone who pulls main with the tree on disk will have it deleted by that commit, and needs `bin/scdata pull live` to get it back.

Tests: `parsed_store_test` 12 tests / 31 assertions; matcher and job suites 25 / 63; `test/loaders/sc_data/` 70 / 194. All green, `standardrb` clean.

🤖


mortik:
**Draft until the production credentials are in place.** Verified against both live hosts:

```
{endpoint: false, bucket: false, access_key_id: false, secret_access_key: false}
```

So the `sc_data_s3` settings are absent in production today. Merging as-is would deploy an image without the parsed tree *and* without the credentials to fetch it — `fetch_parsed_tree!` no-ops silently, the loaders read an empty directory, and `update_in_game_flag` flips every ship to `in_game: false` on the next 22:00 `CheckJob`. `retire_absent` guards the catalogues against that, but the in-game flag has no such guard.

`73002a559` moves the two non-secret values (`SC_DATA_S3__ENDPOINT`, `SC_DATA_S3__BUCKET`) into `env.clear` in `config/deploy.yml`, since `Rails.app.creds` reads ENV before the encrypted credentials. That leaves only the two actual secrets to add:

```
bin/rails credentials:edit --environment production
```

```yaml
sc_data_s3:
  access_key_id: <op://Fleetyards/HETZNER_S3/access_key_id>
  secret_access_key: <op://Fleetyards/HETZNER_S3/secret_access_key>
```

Same key pair `bin/scdata sync` already uses for the raw trees.

Re-verify with the presence check (booleans only, no values):

```
bin/exec-live web -- "bin/rails runner 'puts %i[endpoint bucket access_key_id secret_access_key].to_h { |k| [k,Rails.app.creds.option(:sc_data_s3, k).present?] }.inspect'"
```

Note the two `env.clear` values only appear after a deploy, so expect `endpoint`/`bucket` to stay `false` until this branch ships — the two credential keys are the ones to confirm beforehand.

The payload side is already done: 15,076 objects are live at `parsed/live/`, push and pull both round-tripped.

🤖

mortik:
**Correction to the comment above — no credentials edit is needed after all.**

The production credentials already carry object-storage keys for `config/storage.yml`:

```
s3_protocol: https
s3_endpoint: fsn1.your-objectstorage.com
s3_bucket:   fltyrd-live-storage
```

Same endpoint and same account as the parsed tree's bucket — only the bucket differs (`fltyrd-sc`, where the raw trees already live). Hetzner object-storage keys are project-scoped, so the deployed pair reaches `fltyrd-sc` too.

`c883d586a` falls back to those keys, so production needs only the bucket name — which isn't a secret and landed in `env.clear` in `73002a559`. The `sc_data_s3` keys remain as an override, for a separate account later or for running the CLI outside 1Password.

Net effect: **nothing to add to the credentials file.** One fewer copy of the same key to rotate, and the earlier `{endpoint: false, bucket: false, access_key_id: false, secret_access_key: false}` result no longer blocks this — `endpoint` and `bucket` arrive with the deploy, and the key pair resolves from the existing credentials.

One assumption worth a sanity check before merge: that `fltyrd-sc` and `fltyrd-live-storage` are in the **same Hetzner project**, so the deployed key is authorised for both. If they aren't, the pull raises and the import fails visibly — it fails loud rather than silent — but better to confirm than to find out at 22:00.

Ready for review once that's confirmed.

🤖

PR #4513, opened 2026-08-25
chore(main): release 7.3.0

mortik:
:robot: I have created a release *beep* *boop*
---

### Features

* **api:** serve the fleetchart ship views from their own endpoint ([#4539](https://github.com/fleetyards/fleetyards/issues/4539)) ([9c9441a](https://github.com/fleetyards/fleetyards/commit/9c9441a671c2b10ca75370fd5cea609bb9478a61))
* **cable:** describe every channel with AsyncAPI and broadcast objects ([#4523](https://github.com/fleetyards/fleetyards/issues/4523)) ([1bb301c](https://github.com/fleetyards/fleetyards/commit/1bb301cf875d24f273e7287774131a476785a3b5))
* **cable:** raise on a payload mismatch in tests ([#4525](https://github.com/fleetyards/fleetyards/issues/4525)) ([c09bd90](https://github.com/fleetyards/fleetyards/commit/c09bd90da42558e8500c7d6765ad22789c24d03e))
* **resource-access:** typed enums + grouped-catalog groundwork ([#4213](https://github.com/fleetyards/fleetyards/issues/4213)) ([4487b89](https://github.com/fleetyards/fleetyards/commit/4487b8920e567066327c281436db577f2f719f5f))
* **sc_data:** filter and sort equipment through the build ([#4532](https://github.com/fleetyards/fleetyards/issues/4532)) ([8a80b2e](https://github.com/fleetyards/fleetyards/commit/8a80b2e7c730e9947cbefb0ca97f64c6c7d9820c))
* **sc_data:** give equipment a row per build ([#4522](https://github.com/fleetyards/fleetyards/issues/4522)) ([dbdfba1](https://github.com/fleetyards/fleetyards/commit/dbdfba129d2397f261beef6d9537ecf4066866e0))
* **sc_data:** read commodities from the resource database too ([2712d63](https://github.com/fleetyards/fleetyards/commit/2712d63bc7d1ba5b9a80a8658e5c419c531db00d))
* **sc_data:** read equipment off the build ([#4530](https://github.com/fleetyards/fleetyards/issues/4530)) ([5905238](https://github.com/fleetyards/fleetyards/commit/5905238f13f41442fe52942493b0b2ee80e5713f))
* **uex:** map the three raw materials whose names read backwards ([0f38f4f](https://github.com/fleetyards/fleetyards/commit/0f38f4fb52f63585fb1baf5dd686eb6a950e338d))

### Bug Fixes

[...]

PR #4529, opened 2026-08-26
fix(fleets): rank promote and demote against one ordering

mortik:
Fixes the flaky `FleetsMembersDemoteTest#test_...demotes_the_officer` failure — expected 200, got 400 `You can't demote the last Member of a permanent Fleet Role`. It last showed up on the release PR #4513, which carries no code change, so it is a pre-existing flake on main (2 of the last 20 runs were red, the other one for an unrelated reason).

## Cause

The rendered message points at the wrong branch. `next_fleet_role` and `prev_fleet_role` took their index from `fleet.fleet_roles.ranked` and then read it back out of `fleet.fleet_roles`:

```ruby
index = fleet.fleet_roles.ranked.index(fleet_role)
fleet.fleet_roles[index + 1]
```

`ranked` is lexorank's `where.not(rank: nil).order(rank: :asc)` over a text column; the `has_many` declares no order at all. The two agree only while rank order happens to match insertion order, which is true for a freshly created fleet — hence green almost everywhere. Once they diverge the lookup returns the membership's own role, `promote`/`demote` return nil, and the controller renders 400.

## Reproduction

Adding a fourth role and `move_to(1)`-ing it makes it deterministic:

```
ranked:    [["Admin", "0"], ["Recruit", "00U"], ["Officer", "10"], ["Member", "20"]]
unordered: ["Admin", "Officer", "Member", "Recruit"]
next_fleet_role for Officer: "Officer"   # its own role
```

## Change

Materialise the ranked list once and index only that. Three regression tests cover `next`/`prev`, `promote` and `demote` after a reorder — all three fail without the model change and pass with it.

One caveat: I could not trigger the CI flake itself locally, not even running 70 fleet tests across 4 workers. The ordering bug is proven and sits exactly on the path that produced the 400, so it is very likely the cause of that run, but that part is not proven.

🤖

### DIFF
--- a/app/models/fleet_membership.rb
+++ b/app/models/fleet_membership.rb
@@ -369,19 +369,24 @@ class FleetMembership < ApplicationRecord
   end
 
   def next_fleet_role
-    return if fleet_role.nil?
+    ranked_roles, index = ranked_roles_and_index
+    return if index.nil? || index - 1 < 0
 
-    index = fleet.fleet_roles.ranked.index(fleet_role)
-
-    return if index - 1 < 0
-
-    fleet.fleet_roles[index - 1]
+    ranked_roles[index - 1]
   end
 
   def prev_fleet_role
-    return if fleet_role.nil?
+    ranked_roles, index = ranked_roles_and_index
+    return if index.nil?
 
-    index = fleet.fleet_roles.ranked.index(fleet_role)
-    fleet.fleet_roles[index + 1]
+    ranked_roles[index + 1]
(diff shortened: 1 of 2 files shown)


mortik:
Landed directly on main as eaeee0432 to unblock the branch — same commit content, checks here were all green beforehand. Closing rather than merging because the cherry-pick onto the moved main gave it a new SHA, so GitHub cannot match it automatically.

🤖

PR #4536, opened 2026-08-26
refactor(api): name every array response

mortik:
Top of the stack: #4531 → #4533 → #4534 → this. Review the earlier ones first.

## Why

#4531 left these out on the grounds that only the cardinality was inline and the item type was already named. That was the inconsistent call — the tree already has list components for exactly this shape (`FleetEventAdminsList`, `ModelOptions`, `ManufacturerOptions`), so the same payload was spelled two different ways depending on which endpoint you looked at. This finishes the job.

## What

**29 list components** replacing 87 inline `type: :array` responses. One per item type, shared by every endpoint returning that array — the big ones cover a lot of ground:

| Component | Sites |
| --- | --- |
| `FilterOptionsList` | 21 |
| `PieChartStatsList` | 21 |
| `BarChartStatsList` | 10 |
| `VehicleExportsList`, `ModelsList` | 3 each |
| `FleetInventoryStockItemsList`, `FleetVehicleExportsList`, `HangarInventoryStockItemsList`, `ModelPaintsList`, `PublicVehiclesList` | 2 each |
| 19 more | 1 each |

`FilterOptionsList` and `BarChartStatsList` are in `shared/v1` because both schemas reference them; the rest sit in the narrowest scope that reaches them.

### Naming

Mechanical `<Plural>List` everywhere it works. Six needed a decision, each carrying a comment in the file:

- `ModelsList`, `ImagesList` — `Models` and `Images` already name the **paginated** collections (`items` + `meta`), so the bare array cannot reuse them
- `FleetInvitesList` — same collision with `FleetMembersList`, which is also paginated; named after what `GET /fleets/invites` returns
- `PublicVehiclesList`, `PublicHangarGroupsList`, `ExtendedFleetRolesList` — pluralising the qualifier suffix gives `VehiclePublicsList` and friends, so these read from the resource instead

## Verification

- `oasdiff breaking` on all three specs: **no breaking changes** (oauth: no changes at all). A `$ref` to an array component dereferences to what was there.
- No dangling refs in any spec.
- `validate-schema` / `validate-admin-schema` valid, 67 warnings before and after.
- `pnpm lint:ts` clean — Orval emits plain aliases (`export type FilterOptionsList = FilterOption[]`), structurally identical to the `FilterOption[]` it emitted before, so no call site changes.
- All 87 touched test files: 302 tests, 0 failures.

## After this

No inline request or response types remain anywhere in `test/integration`. The only inline `type: :object` left is the `q` deepObject parameter, where it is required next to `style: :deepObject` with the `$ref` alongside — documented as the one exception.

🤖

### DIFF
--- a/.cursor/rules/backend/api-components.mdc
+++ b/.cursor/rules/backend/api-components.mdc
@@ -35,6 +35,23 @@ An inline body has no name, so Orval invents one from the operation
 up with unrelated TypeScript types that drift, and the frontend has nothing
 stable to import.
 
+An array-only response counts too — it gets a `<Plural>List` component instead
+of an inline `type: :array` around the item `$ref`:
+
+```ruby
+# Good
+schema "$ref": "#/components/schemas/FilterOptionsList"
+
+# Bad
+schema type: :array, items: {"$ref": "#/components/schemas/FilterOption"}
+```
+
+One list component per item type, reused by every endpoint returning it. When
+the plural already names the paginated collection (`Models`, `Images`,
+`FleetMembersList`), the bare array takes the `List` suffix or a name from the
+endpoint — `ModelsList`, `ImagesList`, `FleetInvitesList` — and says so in a
+comment.
+
 ## Placement
 
 - Request bodies → `<scope>/v1/schemas/inputs/<name>_input.rb`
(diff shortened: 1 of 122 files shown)


mortik:
Rebased onto the new base after #4531 and #4533 landed as squashes. Picked up one extra site while doing it: `/models/fleetchart-views` merged in #4539 during review, before the rule was on record, and wrote its response as an inline array — now `FleetchartViewsList`. 🤖

PR #4537, opened 2026-08-26
fix(api): containerFit rejected every request, and the last inline enums

mortik:
Top of the stack: #4531 → #4533 → #4534 → #4536 → this.

Four things, one commit each.

## 1. The three single-value status enums

`HangarSyncSubmitResult.status` (`pending`) and the two cable messages (`failed`, `finished`) were the last inline enums. #4534 argued they are constants rather than enums; that reasoning holds, but it leaves a reader deciding case by case whether an inline enum was deliberate. Three one-value components cost less than that judgement call.

**Written as Ruby class references, not `$ref` strings:**

```ruby
status: V1::Schemas::Enums::HangarSyncSubmitStatusEnum
```

This works — `deep_stringify` in `components/base.rb`, `dsl/context.rb`, `response_context.rb` and `operation_context.rb` all turn a component class into the same `$ref`. A typo is a `NameError` while generating instead of a dangling reference that only `strict_reference_validation` warns about. Roughly 2,100 string refs remain across components and tests; converting them is a mechanical sweep worth its own PR, verifiable by the generated YAML staying byte-identical.

`api-schema-check` on #4536 caught a stale cable document while this was in review: #4534 extracted `ModelHullPartCategoryEnum`, which the cable document embeds through the `Model` schema, without running `bin/generate-asyncapi`. Fixed at the source in #4534 and the stack rebased, so nothing about it lands here.

## 2. The redundant `type: :object` is gone

Every deepObject parameter carried `type: :object` next to its `$ref`. The referenced component is already `type: :object`, so the sibling added nothing — and one endpoint (`public_hangars_embed`) had been written without it all along, producing the same document. Removed from the other 68; oasdiff reports **no changes at all**. The "one inline exception" in the rule is gone with it.

## 3. `containerFit` rejected every request — a real bug

`GET /models` declared the parameter as `additionalProperties: {type: :integer}`. Request validation coerces the strings a query string delivers **only for declared properties**, so:

```
containerFit[16]=2  →  400 {"details":["Invalid query parameter 'containerFit': value at `/16` is not an integer"]}
```

Every container-filtered search 400'd, which means the cargo-grids container filter could not return anything. Nothing covered it — the parameter was declared in the test but never exercised.

`:number` alone does not fix it (`additionalProperties` is the problem, not the type), though `:number` matters too: `q[lengthGteq]=10` validates and an `:integer` in the same position would not. `ContainerFitQuery` spells out the seven sizes `CargoHoldContainerCapacity::CONTAINER_SIZES` validates against:

| request | before | after |
| --- | --- | --- |
| `containerFit[16]=2` | 400 | **200** |
| `containerFit[16]=2&containerFit[32]=1` | 400 | **200** |
| `containerFit[7]=1` (not a standard size) | 400 | 400 |
| `containerFit[16]=x` | 400 | 400 |
| `containerFit[16]=-1` | 400 | 400 |

Two regression tests added. The negative one uses a plain `get` rather than the DSL, because declaring a 400 there would replace the auto-injected `SchemaValidationError` response and move the schema.

`cargo-grids.vue` now annotates its map with the generated `ContainerFitQuery`, so a size the schema does not know is a compile error. Its `CONTAINER_SIZES` literal still duplicates the Ruby constant — the generated artifact is a type, not a runtime array, so there is nothing to import; the comment says so.

## 4. Rule updates

Drops the `q` exception, and records the two things that cost time: query components must declare their properties or validation rejects everything, and numeric filters need `:number` rather than `:integer`. Adds the class-reference form as preferred for new code.

## Verification

- `oasdiff breaking` on all three specs: no breaking changes. Commit 2 alone: no changes at all.
- `bin/generate-asyncapi` + `git diff --exit-code` clean, same for `swagger` — the `api-schema-check` gate.
- `standardrb`, `pnpm lint`, `pnpm lint:ts` clean.
- 752 integration tests across all 69 touched files, plus `test/asyncapi`: 0 failures.

🤖

### DIFF
--- a/.cursor/rules/backend/api-components.mdc
+++ b/.cursor/rules/backend/api-components.mdc
@@ -97,16 +97,38 @@ Check for an existing component before adding one. `SortInput`,
 `StandardError`, `SuccessResponse`, `MessageResponse` and the
 `BaseList` / `Meta` pair cover most small payloads.
 
-## The one inline exception
+## Query parameters
 
-The `q` deepObject query parameter needs a literal `type: :object` alongside
-`style: :deepObject`, and carries its `$ref` to a `<Name>Query` component:
+A deepObject parameter carries a plain `$ref` and nothing else. The
+`type: :object` that used to sit beside it was redundant — removing it produces
+a byte-identical document.
 
 ```ruby
 parameter name: "q", in: :query,
-  schema: {type: :object, "$ref": "#/components/schemas/ModelQuery"},
+  schema: {"$ref": "#/components/schemas/ModelQuery"},
   style: :deepObject, explode: true, required: false
 ```
+
+**Spell the properties out.** Request validation coerces the strings a query
+string delivers only for declared properties, so a map typed through
+`additionalProperties` rejects every request no matter what the value type says
+— `additionalProperties: {type: :integer}` on `containerFit` 400'd every call
+until it was replaced by `ContainerFitQuery`, which lists its seven keys.
+
+Numeric filters use `type: :number`, never `:integer`: only `:number` coerces.
+
+## Reference components by class
+
+`$ref` strings work, but `schema({})`, `parameter`, `request_body` and
+`response ... schema` all accept the component class and emit the same `$ref`:
+
+```ruby
+status: V1::Schemas::Enums::HangarSyncStatusEnum
+```
+
+A typo becomes a `NameError` while generating instead of a dangling reference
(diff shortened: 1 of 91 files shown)


mortik:
Rebased onto the new base. One extra site picked up: the `q` parameter on `/models/fleetchart-views` (merged in #4539 during review) carried the same redundant `type: :object` beside its `$ref`. 🤖

PR #4540, opened 2026-08-26
refactor(api): extract every nested object into a component

mortik:
Top of the stack: #4531 → #4533 → #4534 → #4536 → #4537 → this.

## Why

The last anonymous shapes. 141 objects were nested inside components, and Orval names those after their owner and path, so one shape under several owners became several unrelated types:

```
ComponentControllerPowerRangesHigh   ComponentCoolerPowerRangesHigh
ComponentControllerPowerRangesLow    ComponentCoolerPowerRangesLow
ComponentControllerPowerRangesMedium ComponentCoolerPowerRangesMedium
… and six more component types, for a shape that is two numbers
```

48 generated types for `{start, modifier}`, 24 for `{min, max}`, 16 for the low/medium/high triple. `Model.metrics` held 33 properties inline and got reproduced in full four times over.

## What

**63 components** replacing all 141 sites. Grouped by shape first, so it is far fewer than one per site — the count of distinct shapes was 62, and several turned out to be the same thing under different owners.

Eight component types carried a **byte-identical copy** of `POWER_RANGE_ENTRY` and `POWER_RANGES` as Ruby constants. Those are now two components, and the constants are gone. Shield `resistance` and `absorption` share one `ComponentDamageTypeMap`; weapon `damagePerShot` and `damagePerSecond` share one `ComponentWeaponDamage`; five payloads share `ItemAvailability`; six share `InventoryRef`.

Where two variants differed only in `required`, both are kept — `UserRef` and `UserRefRequired`, `CargoHoldOffset` and `ModelSignatureCrossSection` — so no published contract moves. The comments say why.

## Three things worth reviewing

**A `$ref` cannot be deep-merged into.** openapi-ruby merges a subclass's schema into its parent's, so pointing `Model.links` at a component left `ModelExtended` emitting a `$ref` *beside* the two properties it wanted to add. Same for the admin `Model.media`. Both now use a component that inherits the shared one (`AdminModelMedia < Shared::V1::Schemas::ModelMedia`), which reproduces the old document exactly. I scanned all three specs for `$ref`-with-siblings afterwards: zero.

**A cross-scope subclass inherits the reference.** `ModelModuleMetrics` started in `v1/` and the admin document came out with a dangling `$ref`, because the admin `ModelModule` subclasses the public one. Moved to `shared/v1/`. `bin/generate-schema` failed loudly on this one, but a ref-vs-definition diff over each spec is the reliable check and is now in the rule.

**`AdminModelMedia` had dropped `rsiStoreImage`.** oasdiff classes a removed optional response property as a *warning*, and the admin job runs `--fail-on ERR` — so it passed. The generated client is what caught it. Worth knowing that the admin gate is weaker than the public one.

## Frontend

Only two of the 141 anonymous types were imported by app code (`ModelMetricsHullPartsItem`, `ModelMetricsHullDoorsItem` → `ModelHullPart`, `ModelHullDoor`); everything else is reached through the parent type and keeps working.

One non-obvious consequence: Orval emits an `interface` for a named component and a type alias for an anonymous one, and TypeScript gives interfaces no implicit index signature. Two damage helpers typed `Record<string, unknown>` / `Partial<Record<string, number>>` stopped accepting the value. They are now typed against `ComponentWeaponDamage`, which is the signature they should have had.

## Verification

- `oasdiff breaking` on all three specs, **including warnings**, not just errors: nothing to report.
- No dangling `$ref` and no `$ref`-with-siblings in any spec.
- `bin/generate-schema` + `bin/generate-asyncapi` + `git diff --exit-code`: clean, which is the `api-schema-check` gate.
- `standardrb`, `pnpm lint`, `pnpm lint:ts`: clean.
- 280 integration tests across models, components, equipment, commodities, paints, hardpoints, loadouts, inventories, vehicles, fleet stats, admin models and oauth, plus `test/asyncapi`: 0 failures.
- 141 vitest tests over the touched composables and model components: 0 failures.

Net: −688 lines, and the schemas shrink by about 1,300 lines of duplicated structure.

## After this

Nothing in `app/api_components/` or `test/integration/` defines a request type, a response type, an enum or a nested object inline. What remains inline is only what has nothing to name: free-form `additionalProperties` maps (`Import.input`, `FleetMembersMetrics.membersByRole`, `FleetModelCountsStats`) and `FeatureFlagName`, which is deliberately referenced by nothing.

🤖

### DIFF
--- a/.cursor/rules/backend/api-components.mdc
+++ b/.cursor/rules/backend/api-components.mdc
@@ -72,6 +72,39 @@ the admin components that subclass a v1 one
 (`Admin::V1::Schemas::Models::Model < ::V1::Schemas::Models::Model`) inherit its
 `$ref`s — anything they reach has to be `shared/v1`.
 
+## Nested objects
+
+An object nested inside a component gets its own component. Orval names an
+anonymous one after its owner and path, so the same shape under three owners
+becomes three unrelated types.
+
+```ruby
+# Good
+metrics: Shared::V1::Schemas::ModelMetrics,
+
+# Bad
+metrics: {type: :object, properties: {...}},
+```
+
+Free-form maps (`additionalProperties`, no named properties) stay inline.
+
+Two traps, both learned the hard way:
+
+- openapi-ruby deep-merges a subclass's schema into its parent's, and **a `$ref`
+  cannot be merged into**. Pointing `Model.links` at a component made
+  `ModelExtended` emit a `$ref` next to the properties it wanted to add. The fix
+  is a component that inherits the shared one —
+  `AdminModelMedia < ::Shared::V1::Schemas::ModelMedia`.
+- A cross-scope subclass **inherits the reference**, so anything a v1 component
+  points at must be in `shared/v1/` when an admin component subclasses it. After
+  extracting, diff refs against definitions in each generated spec; a dangling
+  `$ref` only warns.
+
+Also worth knowing: Orval emits an `interface` for a named component and a type
+alias for an anonymous one, and TypeScript gives interfaces no implicit index
+signature. A helper typed `Record<string, unknown>` stops accepting the value —
+type the helper against the component instead.
+
 ## Enum values
 
 Source them from the Rails model when it has them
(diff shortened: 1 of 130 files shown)


mortik:
Rebased onto the new base. One extra site picked up: `FleetchartView.media` from #4539 was nested inline — now `FleetchartViewMedia`. The nested-object count is back to zero. 🤖

PR #4551, opened 2026-08-27
feat(models): show what the game files measure next to the matrix figure

mortik:
`sc_length`, `sc_beam` and `sc_height` have been written by the sc_data loader and read by nothing. The displayed dimensions come from the ship matrix alone, so the fact that the two sources disagree for **91 of 211 ships** has been invisible.

This serves the game-file measurement as `metrics.gameFiles` and shows it under the length, beam and height tiles on the ship page — a quiet second line, and only where it differs by more than 1%. For the other 120 ships nothing changes.

**Fleetcharts and compare are untouched.** They keep reading `metrics.length/beam/height`; this only adds a line to the detail page. That was deliberate: making the difference visible is safe, switching the source is not (see below).

## Why both figures rather than one

A ship has up to three size sets — landed, in flight, and a third for deployable geometry like the Hull C's cargo grid or the Stingray's combat configuration — and **neither source records which one it measured.** Flight is normally the default. So a difference is usually not an error, and it cannot be resolved automatically, only labelled by a person.

The Aurora Mk II is the clearest case: matrix `[27.5, 10.5, 5.5]`, game files `[27.4, 27.7, 8.0]`. The length agrees; the beam nearly triples because the wings are out.

## What the disagreements actually are

Measured, with a tolerance of 5% or 0.5 m — relative alone over-reports small vehicles, where 0.3 m on a 1.5 m bike reads as 25%:

| | |
| --- | --- |
| agree | 120 |
| axis order differs | 36 |
| deployed geometry (length holds, cross-section grows) | 30 |
| unexplained | 25 |
| no game-file measurement (zeros) | 5 |

Deployed geometry has a clear signature: across that group the game figure is larger on beam for 25 ships against 5 smaller, and on height for 32 against 11, while the length mostly agrees.

## Why the axes are not fixed here

The export does not order its axes consistently — the Caterpillar's matrix length of 111.5 m arrives as its game-file height, the X1's length as its beam. That is our bug and the largest group, but it is not fixable from the data we store:

- **`signature_cross_section`** is the only other per-axis figure we keep. Its ordering matches the dimension ordering for **0 of 104** agreeing ships and 2 of 24 reordered ones — it measures something else entirely.
- **"the longest axis is the length"** is wrong for 28 ships. 21 are wider than long (Ares Inferno, Mercury Star Runner, Reliant Mako, the Cyclone family), 7 taller than long (Khartu-Al is 32 m tall, Nox, ATLS, Syulen). Worse, the game files already agree exactly with the matrix for those, so relabelling by size would break 17 currently-correct ships and take the match count from 120 down to 112.
- **Cargo grid geometry** carries dimensions but no positions, and grids are sometimes mounted crosswise to the ship, so it says nothing about orientation either.

The information is in the raw game files, not in the three numbers the parser reduces them to. The fix belongs in the parser, where the entity's orientation is still available.

## Verification

1,300 tests across the models and the whole public API, green. Two new tests cover the served field and the zero case. `lint:ts` exit 0, `standardrb` and prettier clean. The REST schema, the admin schema and the AsyncAPI cable document were all regenerated — the cable document embeds the REST component and rejects an undeclared property, which is how the first run failed.

🤖

### DIFF
--- a/app/api_components/shared/v1/schemas/model_metrics.rb
+++ b/app/api_components/shared/v1/schemas/model_metrics.rb
@@ -27,6 +27,18 @@ module Shared
             extendedFleetchartOffsetBeam: {type: :number},
             height: {type: :number},
             heightLabel: {type: :string},
+            # What the game files measure, where the fields above are what the
+            # matrix publishes. Absent when the export has no measurement, and
+            # the two disagree often enough that neither replaces the other.
+            gameFiles: {
+              type: :object,
+              properties: {
+                length: {type: :number},
+                beam: {type: :number},
+                height: {type: :number}
+              },
+              additionalProperties: false
+            },
             hydrogenFuelTankSize: {type: :number},
             isGroundVehicle: {type: :boolean},
             length: {type: :number},
--- a/app/frontend/frontend/components/Models/BaseMetrics/index.vue
+++ b/app/frontend/frontend/components/Models/BaseMetrics/index.vue
@@ -26,19 +26,21 @@ const props = withDefaults(defineProps<Props>(), {
 const soldAt = computed(() => props.model.availability.soldAt);
 const rentalAt = computed(() => props.model.availability.rentalAt);
 
-// The game files measure a ship themselves, and for about a third of ships their
-// figure disagrees with the one the matrix publishes -- by anything from a third
-// under to nearly triple. Neither source is reliably right, so the second figure
-// is shown rather than resolved away, and only where it actually differs: for the
-// other two thirds it would just repeat the number above it.
+// The game files measure a ship themselves, and for 91 of 211 ships their figure
+// differs from the one the matrix publishes. Often that is not an error: a ship
+// has up to three size sets -- landed, in flight, and with a cargo grid or wings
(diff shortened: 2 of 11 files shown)


mortik:
Closing — wrong call on my part.

Fleetyards should show the one correct figure for a ship, not surface CIG's inconsistencies or our own parser's mistakes to visitors. Showing both numbers side by side does exactly that: a reader sees two dimensions and no way to tell which to believe, and about a third of what they would be seeing is our axis-mapping bug rather than anything real about the ship.

The investigation behind this stands and is worth keeping:

- 91 of 211 ships have a game-file measurement that differs from the matrix figure
- ~36 of those are our axis mapping — `maxBoundingBoxSize` does not use a consistent convention, so `length <- metrics.y` is wrong for a fifth of ships
- ~30 are configuration differences, and neither source records which configuration it measured
- the orientation cannot be derived from the current export; that is tracked where the export lives

So the right order is the reverse of what this PR assumed: get to one trustworthy figure first, then display it. Nothing to show until then.

If the disagreement needs to be reviewable while that is sorted out, the place for it is the admin area, not the public ship page.

🤖

PR #4608, opened 2026-08-30
fix(forms): stop printing the validation error next to the checkbox label

mortik:
`FormCheckbox` and `FormToggle` rendered a bare `{{ errorMessage }}` text node straight after the `</label>`, so an invalid control read as:

> Accept terms**The terms field is required**

— the message unstyled and glued to its own label.

## Why removal rather than restyling

Both components already signal the error the way the rest of the design system does, and all three paths stay in place:

- `v-tooltip.right="errorMessage"` on the input
- `aria-invalid` on the input
- the `--with-error` class that turns the label red

The SCSS in each component implements that red-label state deliberately, with a comment saying it matches `base-input--with-error` on the fields that already did it — and there is no CSS anywhere for the stray text node. `FormInput`, `FormTextarea` and `FormDatePicker` render no such node; they rely on the tooltip alone.

So the red label in the screenshot was the *correct* styling working. The white text after it was the leftover.

## Scope

`FormToggle` had the identical node and is fixed too — it just sits in a later row on the visual-tests page, so it is less obvious.

The node has been there since the Vue 3 migration (#2655) and was never styled. Nothing asserts on it: no unit or e2e test references the rendered message. Prettier and eslint pass on both files.

## Not addressed here

- **Tooltip placement.** On a narrow grid column the `v-tooltip.right` box overlaps the field to its right. Pre-existing and unrelated to this node.
- **Error accessibility.** With the node gone the message lives only in a tooltip plus `aria-invalid`, which is the existing house pattern but is weak for screen readers. Giving every control a real `aria-describedby` error element would be a design-system-wide change, so it is deliberately out of scope for a bug fix.

🤖

### DIFF
--- a/app/frontend/shared/components/base/FormCheckbox/index.vue
+++ b/app/frontend/shared/components/base/FormCheckbox/index.vue
@@ -145,7 +145,6 @@ const innerPlaceholder = computed(() => {
     <label :for="uuid">
       {{ innerLabel }}
     </label>
-    {{ errorMessage }}
   </div>
 </template>
 
--- a/app/frontend/shared/components/base/FormToggle/index.vue
+++ b/app/frontend/shared/components/base/FormToggle/index.vue
@@ -118,7 +118,6 @@ const innerPlaceholder = computed(() => {
         {{ innerLabel }}
       </span>
     </label>
-    {{ errorMessage }}
   </div>
 </template>
 


mortik:
Closing in favour of the forms work on `feat/form-redesign`, which has since made the change this PR deliberately scoped out.

This was the right call for a bug fix. The stray node was unstyled and unreferenced, and the note here says plainly what removing it leaves behind:

> With the node gone the message lives only in a tooltip plus `aria-invalid`, which is the existing house pattern but is weak for screen readers. Giving every control a real `aria-describedby` error element would be a design-system-wide change, so it is deliberately out of scope for a bug fix.

That design-system-wide change is now done. All seven controls — `FormInput`, `FormTextarea`, `FormCheckbox`, `FormToggle`, `RadioList`, `FormDatePicker`, `FormDateTime` — render the message as a real element below the control, tied to the input with `aria-describedby` and marked `role="alert"`, with its line reserved so validating cannot move the page.

So the two disagree about more than the node. Two points where this PR's reasoning has been overtaken:

**The tooltip is not a viable home for the message.** It cannot be triggered by touch at all, so on a phone the message is invisible; on a desktop it takes a hover over the offending field to find out why the form refused you. The tooltip now survives only on login, signup, request-password and change-password — the four compact forms where the label folds away and there is no line to give a message — and even there it repeats what is already in the DOM.

**The red label was not the correct styling working.** The signature — a field's cap, a checkbox's edge — already says invalid, so colouring the label as well says it twice and colours the half that needs to stay readable. The message carries the colour now and the label does not.

Both files this touches are edited on the same lines by that branch, so leaving this open means one of the two has to be reverted later. The symptom is fixed either way: nothing renders "Accept termsThe terms field is required" any more.

🤖

PR #4624, opened 2026-08-31
feat(discord): answer slash commands from an interactions endpoint

mortik:
> Stacked on #4623 — the base ref is `fix/discord-install-permissions`, not `main`. Review the four commits from `docs(discord)` onwards; the first commit belongs to the other PR.

Fleetyards' Discord presence only ever pushed: five webhook posters and a one-way event sync. Nobody in a server could ask it anything. This is the foundation that changes that, plus one command so the foundation is testable end to end.

`docs/exec-plans/discord-bot.md` plans the whole programme in six phases. Two findings from reading the tree shaped it:

- **`lib/discord/scheduled_event_rsvp_handler.rb` is complete and unreachable.** It maps a Discord user to a Fleetyards user, finds the `FleetEvent`, creates a slot-less `interested` signup and carefully refuses to downgrade a stronger commitment made on Fleetyards. Nothing calls it: it was written for `GUILD_SCHEDULED_EVENT_USER_ADD`, a **Gateway** event, and the only `discordrb` require in the tree is `discordrb/webhooks`. So "Interested" in Discord never arrives.
- **HTTP interactions cannot carry those events either.** Phase 3 therefore polls `GET /guilds/{id}/scheduled-events/{id}/users` from the existing sync rather than adding a Gateway process — which also survives restarts and missed windows, where a Gateway drops whatever happens while it is down.

## Why an HTTP endpoint and not a Gateway bot

Discord POSTs every interaction to an Interactions Endpoint URL, so slash commands are a route, a controller and a Sidekiq job — all patterns this repo has — instead of a long-running supervised process it has no pattern for under Kamal. What that gives up is Gateway-only events; only phase 3 wanted one, and it has another route.

Consequences that shaped the code:

- **Discord's 3 second deadline is hard and unrecoverable**, so no command runs inline. Every one is acknowledged with a deferred response and answered from the job. "Fast" locally still means a cold connection pool under load.
- The follow-up is authenticated by the interaction token in the URL, not the bot token, so it gets **its own client**: `ApiClient` refuses to run without a bot token, and that guard is worth keeping rather than relaxing.
- Command jobs get **their own Sidekiq queue, listed first** — it is the only queue with a user watching a spinner, and a loader burst on `notifications` would read as a bot that takes half a minute to answer.

## The security boundary

The endpoint is public, session-less and unauthenticated; the Ed25519 signature check is all there is.

- Verifies `request.raw_post`, because Rails' parsed and re-serialised params are a different byte sequence and would never match.
- `OpenSSL::PKey.new_raw_public_key` rather than the `ed25519` gem. The gem *is* in `Gemfile.lock`, but only as a transitive dependency of the SSH stack Kamal uses — a `bundle update` that drops it would silently take this boundary with it. OpenSSL 3.6 is already a hard requirement.
- Rejects timestamps outside a five minute window **in both directions**, so a captured request is not replayable forever and a far-future timestamp cannot buy itself a longer life.
- Unverified requests get a **401**, not a 200 with an error body: Discord probes a newly saved endpoint URL with a deliberately invalid signature and refuses the URL unless the probe is rejected.

Two placement decisions worth flagging in review:

- The route sits next to `/emails/inbound` in `config/routes.rb`, deliberately **outside** `:api_routes`. Everything under `api/v1` is public API surface with hand-written OpenAPI components; a Discord RPC endpoint whose body Discord defines must not reach the generated schema or the TS client. The integration test lives outside `test/integration/api/v1` for the same reason.
- The controller inherits `ActionController::Base`, **not** `ApplicationController` — the latter adds HTTP basic auth whenever `basic_auth.password` is set, which is how staging is protected, and Discord cannot supply those credentials. Every interaction would 401 there.

## `/ship`

Reuses the site's own matcher — the `search_cont` ransack alias over name, slug and manufacturer slug — so the bot and the ship list agree on what a name matches, and `Model.visible.active` so it cannot answer with models the website itself hides. An exact name wins over a longer partial match; an ambiguous one lists candidates instead of guessing.

Discord sends the invoking member's own client locale on every interaction, so answers render under `I18n.with_locale` and the copy is in all seven `discord.yml` files. `Discord::Locale` matches exactly first, so `zh-CN` and `zh-TW` stay distinct rather than both collapsing to `zh`.

Registration is a full replacement of the global command list, which makes `discord:commands:sync` idempotent. It stays a task rather than a deploy hook: the endpoint is rate-limited per day and definitions change far less often than deploys.

## Verified

- `bin/rails test test/lib/discord test/jobs/discord test/integration/discord_interactions_test.rb` — **94 runs, 157 assertions, 0 failures** (57 new; the pre-existing sync and RSVP tests still pass)
- `bundle exec standardrb` on every new file — clean
- `bin/rails zeitwerk:check` — clean
- `bin/feature-flags validate` — valid, 16 flags
- All seven locale files parse

The signature verifier is tested with a real Ed25519 keypair rather than a stubbed verifier: a genuine signature verifies, and a tampered body, a foreign key, a shifted timestamp, a stale one, a future one and malformed hex do not.

## To switch on after merge

1. `DISCORD__PUBLIC_KEY` in credentials — the endpoint rejects everything while it is blank
2. Interactions Endpoint URL in the portal → `https://fleetyards.net/discord/interactions`
3. `bin/rails discord:commands:sync`
4. Flag `discord_commands` in `/admin/features`

The flag gates dispatch only; PING always answers, so step 2 works before step 4.

## Not in this PR

Phases 2–6: the remaining read-only commands, the RSVP round trip, event reminders with slot counts, wishlist DMs, and fleet role sync. Phase 6 moves the install permissions again (Manage Roles, `1 << 28`) and will need existing servers to re-authorise, since Discord does not upgrade an existing grant.

Also noticed: `refs/heads/feature/discord-bot` on origin is an abandoned WIP branch from the Rails 6.1 era attempting the same thing. Left untouched — this branch is named `feature/discord-interactions` to avoid it.

🤖

### DIFF
--- a/.env.tpl
+++ b/.env.tpl
@@ -8,6 +8,7 @@ SC_DATA_S3__BUCKET=fltyrd-sc
 SC_DATA_S3__ACCESS_KEY_ID=op://Fleetyards/HETZNER_S3/access_key_id
 SC_DATA_S3__SECRET_ACCESS_KEY=op://Fleetyards/HETZNER_S3/secret_access_key
 # DISCORD__OAUTH_SECRET=op://Fleetyards/DISCORD_OAUTH/credential
+# DISCORD__PUBLIC_KEY=op://Fleetyards/DISCORD_BOT/public_key
 GITHUB__OAUTH_CLIENT_ID=op://Fleetyards/GITHUB_OAUTH/client_id
 GITHUB__OAUTH_SECRET=op://Fleetyards/GITHUB_OAUTH/credential
 TWITCH__OAUTH_CLIENT_ID=op://Fleetyards/TWITCH_OAUTH/client_id
--- a/app/controllers/discord/interactions_controller.rb
+++ b/app/controllers/discord/interactions_controller.rb
@@ -1,0 +1,114 @@
+# frozen_string_literal: true
+
+module Discord
+  # Discord's Interactions Endpoint. Every slash command arrives here as a
+  # POST; there is no session, no cookie and no user.
+  #
+  # Inherits ActionController::Base rather than ApplicationController on
+  # purpose: ApplicationController adds HTTP basic auth whenever
+  # basic_auth.password is set, which is how staging is protected. Discord
+  # cannot supply those credentials, so every interaction would 401 there.
+  class InteractionsController < ActionController::Base
+    skip_forgery_protection
+
+    # Interaction request types.
+    PING = 1
+    APPLICATION_COMMAND = 2
+
+    # Interaction response types.
+    PONG = 1
+    MESSAGE = 4
+    DEFERRED_MESSAGE = 5
+
+    def create
+      return head :unauthorized unless verified?
+
+      case payload["type"]
+      # PING answers regardless of the feature flag: saving the endpoint URL in
+      # the Discord portal is what triggers it, and that has to be possible
+      # before the commands are switched on.
+      when PING then render json: {type: PONG}
+      when APPLICATION_COMMAND
+        Flipper.enabled?(:discord_commands) ? acknowledge_command : refuse_command
+      else head :no_content
+      end
+    end
+
+    # Discord probes a newly saved endpoint URL with a deliberately invalid
+    # signature and refuses to accept the URL unless that probe is rejected --
+    # so an unverified request must be a 401, never a 200 with an error body.
+    private def verified?
+      return false unless SignatureVerifier.configured?
+
+      SignatureVerifier.new.valid?(
+        signature: request.headers[SignatureVerifier::SIGNATURE_HEADER],
+        timestamp: request.headers[SignatureVerifier::TIMESTAMP_HEADER],
+        body: raw_body
+      )
+    end
+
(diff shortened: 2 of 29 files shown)


mortik:
CI caught something the local run could not: adding `discord_commands` to `config/feature_flags.yml` changes the **public** API schema, because `FeatureFlagName` is a generated enum over the registry. `api-schema-check` failed on the uncommitted `swagger/v1/schema.yaml`.

Regenerated with `./bin/generate-schema` and folded into the commit that adds the flag — the diff is exactly the two lines CI wanted:

```
     FeatureFlagName:
       enum:
+      - discord_commands
       x-enumNames:
+      - DISCORD_COMMANDS
```

The generated TS clients under `app/frontend/services/` are gitignored, so nothing else needed committing. Worth remembering for later phases: every new flag is an API schema change here, even when no endpoint moves.

The branch was force-pushed (amend, not a new commit), and #4625 has been rebased onto the new tip. 🤖

PR #4632, opened 2026-08-31
feat(discord): sync fleet membership to Discord roles

mortik:
> Stacked on #4631 → #4628 → #4627 → #4625 → #4624 → #4623. The base ref is `feature/discord-wishlist-dms` — review the two commits from `feat(discord): request Manage Roles on install` onwards.

The last phase, and the one org admins actually ask for: an accepted member gets the fleet's verified-member role in Discord and the role mapped to their rank, and loses both when they leave.

## The invariant worth reviewing first

`MemberRoleSync` may only add or remove role ids **the fleet itself configured**. `managed_role_ids` is that entire universe:

```ruby
[setting.discord_member_role_id] + fleet.fleet_roles.pluck(:discord_role_id)
```

A role handed out by hand in Discord — a colour, a game, a moderator role — is invisible to this code and can never be taken away by it. There is a test named after exactly that:

```
never removes a role the fleet did not put under our control
```

An accepted member gets the roles; invited, requested, declined and discarded members get none, which is what makes leaving a fleet take them off. A member who is not in the Discord server at all is not a failure — plenty of fleet members never join it.

## Only two changes trigger a sync

State and rank. Ship filters, hangar groups and the primary flag have no bearing on roles, and enqueuing on every membership save would put a Discord round trip behind ordinary edits. `discord_roles_affected?` is one line and there is a test that an unrelated edit enqueues nothing.

## The permission change, and telling fleets about it

The install mask moves to `8_859_419_648` (adding Manage Roles, `1 << 28`). **Discord does not upgrade a server that installed under the old mask** — it keeps its original grant — so the constant alone changes nothing for existing installs.

`Discord::RoleCapability` is what closes that gap, running from the discord-status endpoint the fleet settings page already polls:

| Code | Meaning | Fixable by |
| --- | --- | --- |
| `missing_manage_roles` | installed before Manage Roles was requested | re-authorising the bot |
| `role_above_bot` | target role at or above the bot's highest role (names the role) | the server admin, not us |
| `unknown_role` | a mapping whose Discord role was deleted | fixing the mapping |
| `ok` | — | — |

It reports **nothing at all** until the fleet has actually mapped a role, so no one is told to re-authorise for a feature they are not using. Permissions are read as the union of the bot's roles, and Administrator counts as Manage Roles.

A 403 on a role change is reported, not retried: neither a missing permission nor a role hierarchy changes by trying again. Rate limits and 5xx still retry.

## What is configurable, and what is not — please read

`fleet_notification_settings.discord_member_role_id` has an update endpoint, an input schema entry and a response field, so **the verified-member role works end to end today**.

`fleet_roles.discord_role_id` does **not**. `FleetRole` has no write endpoint anywhere in the app — the public API exposes `index` only, and not even `resource_access` is writable through it. The rank mapping is fully implemented and tested, but until fleet roles become editable it can only be set from a console.

I did not build that endpoint here: it needs its own policy, schema and tests, and it is a fleet-roles feature rather than a Discord one. Flagging it rather than shipping it quietly, and rather than leaving the rank mapping out and needing a second migration next to it later. **If you would rather nothing inert shipped, say so and I will drop the column and the mapping from this PR.**

Likewise not included: a settings UI for picking which Discord role to use. The API accepts the id; the fleet settings page has no field for it yet.

## Verified

- `bin/rails test` over the Discord, notification, membership and fleet-role suites — **273 runs, 569 assertions, 0 failures** (30 new)
- The existing membership suites still pass with the new `after_commit` in place: `fleet_membership_test`, `fleet_membership_discard_test`, `fleets_members_promote_test`, `fleets_members_demote_test`, `fleets_membership_accept_invitation_test` — 30 runs, 0 failures
- `bundle exec standardrb` on every touched file — clean
- `./bin/generate-schema` — `rolesOk` / `rolesCode` / `rolesDetail` on the status component, `discordMemberRoleId` on both the setting response and `FleetNotificationSettingUpdateInput`

## The stack, for reference

| PR | Phase |
| --- | --- |
| #4623 | install permissions ↔ portal |
| #4624 | interactions endpoint, signature verification, `/ship` |
| #4625 | event RSVPs from Discord |
| #4627 | `/loaner`, `/compare`, `/hangar`, `/fleet` |
| #4628 | event reminders with slot counts |
| #4631 | notifications as Discord DMs |
| this | fleet membership → Discord roles |

🤖

### DIFF
--- a/app/api_components/v1/schemas/fleets/fleet_notification_discord_status.rb
+++ b/app/api_components/v1/schemas/fleets/fleet_notification_discord_status.rb
@@ -15,7 +15,10 @@ module V1
             guildId: {type: :string},
             guildName: {type: :string},
             status: {type: :integer},
-            installUrl: {type: :string}
+            installUrl: {type: :string},
+            rolesOk: {type: :boolean},
+            rolesCode: {type: :string},
+            rolesDetail: {type: :string}
           }
         })
       end
--- a/app/api_components/v1/schemas/fleets/fleet_notification_setting.rb
+++ b/app/api_components/v1/schemas/fleets/fleet_notification_setting.rb
@@ -17,6 +17,7 @@ module V1
             },
             discordGuildId: {type: :string},
             discordChannelId: {type: :string},
+            discordMemberRoleId: {type: :string},
             # discord_webhook_url is encrypted; never returned, only writable.
             discordWebhookConfigured: {type: :boolean}
           },
--- a/app/api_components/v1/schemas/inputs/fleet_notification_setting_update_input.rb
+++ b/app/api_components/v1/schemas/inputs/fleet_notification_setting_update_input.rb
@@ -11,6 +11,7 @@ module V1
           properties: {
             discordGuildId: {type: [:string, :null]},
             discordChannelId: {type: [:string, :null]},
+            discordMemberRoleId: {type: [:string, :null]},
             discordWebhookUrl: {type: [:string, :null]},
             enabledInAppEvents: {type: :array, items: {type: :string}}
           }
--- a/app/controllers/api/v1/fleet_notification_settings_controller.rb
+++ b/app/controllers/api/v1/fleet_notification_settings_controller.rb
@@ -38,7 +38,7 @@ module Api
 
         begin
           guild = ::Discord::ApiClient.new.get_guild(@setting.discord_guild_id)
-          {ok: true, guildId: guild["id"], guildName: guild["name"]}
+          {ok: true, guildId: guild["id"], guildName: guild["name"]}.merge(roles_payload)
         rescue ::Discord::ApiClient::Error => e
           code = case e.status
           when 401 then "invalid_token"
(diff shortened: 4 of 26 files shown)


mortik:
Added the backfill — without it the role mapping only ever applies to members whose state or rank changes *after* configuration, which is the opposite of what a rollout needs.

Three cases now enqueue one:

| Trigger | Job | Scope |
|---|---|---|
| `discord_member_role_id` set | `BackfillFleetMemberRolesJob` | the whole fleet |
| a `FleetRole.discord_role_id` set | `BackfillFleetMemberRolesJob(fleet_id, role_id)` | that rank only — all a single mapping can affect |
| a Discord `OmniauthConnection` created | `BackfillUserMemberRolesJob` | that member's accepted memberships |

Both jobs consider only **accepted members who actually linked Discord**: for anyone else the per-member sync can only decide to do nothing, so enqueuing it is waste. The fleet backfill spreads its work over time, since one membership costs a `get_guild_member` plus a call per role change and a large fleet would otherwise arrive at Discord in one burst.

15 new tests, including that an unrelated setting change (`discord_channel_id`) enqueues nothing and that linking a *non*-Discord provider enqueues nothing. Full run over the role, membership, job and status suites: **226 runs, 424 assertions, 0 failures**. `standardrb` clean.

**Still not handled, deliberately:** unlinking a Discord account leaves the roles in place. The uid the sync needs disappears with the record, and revoking needs a "remove every managed role from this uid" operation rather than the desired/current diff — the membership is still accepted, so the diff would want to add them back. That is a separate change; say the word and I will write it. 🤖

PR #4635, opened 2026-08-31
fix(forms): make the affix select a real control

mortik:
Found while answering "what do we do with `suffix-select`" ahead of the
FilterGroup rename.

Two logistics quantity fields put a bare `<select class="base-input__suffix-select">`
into `FormInput`'s suffix slot for their unit picker. **That class has no rule
anywhere in the codebase**, so both have been rendering as raw platform controls
inside a field that has none — a `<select>` inherits neither font nor colour,
which is the same trap the `FilterGroup` trigger hit as a `<button>`. The `<span>`
sibling beside it looks right only because the affix wrapper colours its text
children by inheritance.

## Why it is a component and not a rule

Styling it from `FormInput` is the obvious repair and does not work. The element
arrives through a slot, so it carries the *caller's* scope attribute, not
`FormInput`'s. `:slotted` answers that in principle, but this project keeps
component styles in an imported file and the runtime marker comes from analysing
the `<style>` block itself — which sees only the `@import`.

Verified rather than assumed: with `:slotted` in place the compiled CSS did carry
`[data-v-…-s]`, and the element in the DOM had **no scope attribute at all**, so
the rule could never match. No other component in the codebase uses `:slotted`,
so there was no precedent to check against.

A component with its own scope sidesteps the whole question, and removes the
duplicated markup at both call sites. Inheritance still reaches it from the affix
wrapper in the DOM.

## Verified

Live on a local server: `appearance` reset, the field's own font, the affix's
colour, a caret, no border, transparent background, three options, and its
accessible name from `labels.logistics.unit` — which already exists in all seven
locales.

It is also now on the visual-tests forms page, where an affix select had no
coverage at all.

🤖

### DIFF
--- a/app/frontend/frontend/components/Fleets/Logistics/InventoryItemModal/index.vue
+++ b/app/frontend/frontend/components/Fleets/Logistics/InventoryItemModal/index.vue
@@ -10,6 +10,7 @@ import Modal from "@/shared/components/AppModal/Inner/index.vue";
 import Btn from "@/shared/components/base/Btn/index.vue";
 import { BtnSizesEnum } from "@/shared/components/base/Btn/types";
 import FormInput from "@/shared/components/base/FormInput/index.vue";
+import AffixSelect from "@/shared/components/base/FormInput/AffixSelect/index.vue";
 import { InputTypesEnum } from "@/shared/components/base/FormInput/types";
 import FormTextarea from "@/shared/components/base/FormTextarea/index.vue";
 import FormFileInput from "@/shared/components/base/FormFileInput/index.vue";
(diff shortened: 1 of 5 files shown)


mortik:
Closing in favour of #4639, which does this as a variant of `BaseSelect` rather
than a separate component.

The challenge was fair: this wrapped a native `<select>`, which opens the
platform's own dropdown. The app would have ended up with two selects that
diverge the moment either is opened — different popover, different keyboard
handling, no combobox semantics on one of them. That is the second select language
the `FilterGroup` → `BaseSelect` rename exists to remove, reintroduced one
component later.

Everything worth keeping from here is in #4639: both call sites lose their
hand-rolled `<select class="base-input__suffix-select">` and the class that had no
rule behind it, and the affix select gets the visual-tests coverage it never had.
The accessible-name fix went with it and grew — the trigger turned out to be
unnamed at 27 `no-label` call sites and in every `searchable` select, not just in
the affix.

🤖

PR #4642, opened 2026-08-31
fix(fleets): accept the s sort param on the member list

### LINKED ISSUE #4637: Server Error Members List
Going to members, using sorting results in a Server Error
https://fleetyards.net/fleets/orison/members/?s=username%20asc
https://fleetyards.net/fleets/orison/members/?s=rsiHandle%20asc
https://fleetyards.net/fleets/orison/members/?s=acceptedAt%20asc
https://fleetyards.net/fleets/orison/members/?s=lastActiveAt%20asc
https://fleetyards.net/fleets/orison/members/?s=username%20asc

Image

Image

Image

mortik:
Sorting the fleet member list returned a **400**, not the 500 the report suggested — `FilteredList` renders any non-403 error as the generic ServerError page, which is why it read as a server error.

```
q[s]=username asc     -> 400 Request validation failed:
                             object property at `/s` is a disallowed additional property
q[sorts]=username asc -> 200
```

The table header writes the active sort into the URL as `?s=<field> <dir>`, and `useFilters.getQuery()` forwards the whole route query as `q`. `FleetMemberQuery` declared only `sorts` next to `additionalProperties: false`, so openapi-ruby request validation rejected the request before the controller ran. `410fc1971` established the fix but only applied it to three schemas.

## What's in here

- **The member list** — `FleetMemberQuery` declares `s`, the concern permits it, the controller normalizes it into `sorts`. Regression tests for all four sortable columns, plus one asserting an explicit `sorts` wins over `s`.
- **Every other sortable endpoint** — of 25 schemas that accept `sorts`, only 3 accepted `s`. Now all of them do, and the controllers behind them permit and normalize it. The same 400 was waiting on the fleet and hangar inventory logs, the ship cargo page, and the admin model and vehicle lists.
- **`normalize_sort_params`** — it deleted `s` only when `s` was the key that filled `sorts` in. With both present the `||=` short-circuited, `s` survived into ransack, and since ransack reads `s` itself the unvalidated value outranked the whitelisted `sorts`. Caught by the new precedence test.
- **A live 500 on `/fleets/:slug/stats/members`** — the role breakdown is a grouped count, so any sort param left an ungrouped column in the ORDER BY and Postgres rejected the query (`PG::GroupingError`). `sorts` alone was enough to trigger it before this change. Both stats controllers now strip the sort keys the way `model_counts` already did.
- The admin vehicle list permitted neither `s` nor `sorts`, so its sort never reached ransack. Fixed too.

## Why not fix it in `useFilters`

Translating `s` to `sorts` inside `getQuery()` would be a one-liner covering every page — but `getQuery()` also builds the browser URL via `debouncedFilter` → `router.replace`. Renaming the key there rewrites the URL to `?sorts=…`, which `useTableSorting` no longer recognises, breaking the sort indicator and every bookmarked sort link. The URL contract stays `s`; the API learns it.

## Verification

- `fleets_members_index`, `fleets_stats_members`, `public_fleets_stats_members` — 21 tests green, including the four new column sorts and the precedence case
- 127 tests across the touched v1 endpoints — green
- 358 tests across the touched admin endpoints — green
- 58 tests across filters and missions — green
- `standardrb` and `pnpm lint:ts` clean
- Schema regenerated; the diff is 116 lines of additive optional properties, so `api-schema-breaking` should stay green

Plan: `docs/exec-plans/4637-members-list-sorting-server-error.md`

Resolves #4637

🤖

### DIFF
--- a/app/api_components/admin/v1/schemas/queries/admin_fleet_member_query.rb
+++ b/app/api_components/admin/v1/schemas/queries/admin_fleet_member_query.rb
@@ -13,6 +13,9 @@ module Admin
               usernameCont: {type: :string},
               stateEq: {type: :string},
               roleCont: {type: :string},
+              s: {anyOf: [{
+                type: :array, items: ::Shared::V1::Schemas::Sorts::FleetMembershipSortEnum
+              }, ::Shared::V1::Schemas::Sorts::FleetMembershipSortEnum]},
               sorts: {anyOf: [{
                 type: :array, items: ::Shared::V1::Schemas::Sorts::FleetMembershipSortEnum
               }, ::Shared::V1::Schemas::Sorts::FleetMembershipSortEnum]}
--- a/app/api_components/admin/v1/schemas/queries/admin_notification_query.rb
+++ b/app/api_components/admin/v1/schemas/queries/admin_notification_query.rb
@@ -15,6 +15,9 @@ module Admin
               readAtNull: {type: :boolean},
               archivedAtNull: {type: :boolean},
               searchCont: {type: :string},
+              s: {anyOf: [{
+                type: :array, items: ::Admin::V1::Schemas::Sorts::AdminNotificationSortEnum
+              }, ::Admin::V1::Schemas::Sorts::AdminNotificationSortEnum]},
               sorts: {anyOf: [{
                 type: :array, items: ::Admin::V1::Schemas::Sorts::AdminNotificationSortEnum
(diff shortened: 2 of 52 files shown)


iNgeon:
Still getting server error
https://github.com/fleetyards/fleetyards/issues/4637

mortik:
It will be in the next release hopefully this evening CET

PR #4695, opened 2026-09-03
fix(notifications): stand the placeholders in the frame the records land in

mortik:
Placeholder rows now stand where the records land — on the notification pages, and in every list built on `ListGroup`.

## The notification jump

The waiting state and the loaded page disagreed on five measurements, so the list moved twice on arrival: once sideways, once down. Measured on the page at 1280px with the answer held back, before → after:

| | before | after |
|---|---|---|
| list column | full width → 361px | 361px both |
| reading pane | absent while waiting | 567×240 both |
| row icon (checkbox) | 365 → 400 | 365 both |
| row title (icon footprint) | 395 → 400 | 400 both |
| selection bar | 34px → 51px | 51px both |
| heading count line | 32px → 72px | 72px both |

The skeleton slot draws the frame now: the real selection bar (disabled, nothing to tick) and the real reading pane in the state it is about to be in, with the placeholders in the list's own column and a checkbox on every row. `ListSkeleton` is built to the row's two boxes rather than to their insets added together, so the checkbox, the icon and both lines sit where the record's own will.

An empty answer used to replace all of it — bar, list and pane together — with the empty box. The box goes in the list's own column instead, so an empty inbox is still a two-pane page.

## The measurement that never landed

A list's placeholders reserve what one of its rows measured last time, and this loop was never closing on these pages: `useReportListGeometry` reaches the geometry through inject, and the geometry is provided by the `FilteredList` the page renders — inject only looks upwards. Every load reserved a single-line row (72px) against rows that run to two (99px). Naming the list files the reading under the same key the frame reads back; second load now reserves 99px against a 99.375px row.

## Elsewhere

- **`ListGroup`** answered a load with a spinner over an empty box. It draws placeholder rows now, built to the row's own recipe: 47px against 47px for a pill-and-name row, 61px against 61px once it carries buttons. `InlineEditableList` passes the count through.
- **The selection bar on a phone**: once the bar wraps, the checkbox was centred against a two-line bar — 26.5px below the count and level with the buttons. It is held in one box with the count now.
- **The clear button is gone.** It only did something the checkbox could not in one state, and a minus box clears the selection in one click instead. Ticking the whole page from a partial selection is the two-click move now, which is the rarer one. Each of the three states says what a click will do; a disabled box says nothing.
- **The empty box** was 600px wide whatever it stood in, which ran it and its headline over the reading pane. Both box widths are ceilings now. It also offered to reset filters that were never applied — any query parameter counted as one, so the tab and the sort order read as filters.

## Checks

- New: 9 `ListGroup`, 2 `InlineEditableList`, 2 `ListSkeleton`, 6 `useListGeometry` tests. 189 shared-component and composable tests green, `lint:ts` clean.
- Verified against the running app rather than by eye: the numbers above come from measuring both states of the real page with the API held back, and the empty state from deleting every notification (bar and pane in place, box inside its column, no reset actions).
- New visual-test cases: lists → the loading list group beside the list it becomes; notification centre → selectable placeholders beside the rows they stand in for, and the selection bar in its three states plus a phone-width column; states → the empty box in a narrow column.

🤖

### DIFF
--- a/app/frontend/admin/components/Notifications/ListItem/index.vue
+++ b/app/frontend/admin/components/Notifications/ListItem/index.vue
@@ -220,9 +220,16 @@ defineExpose({ focus: () => select.value?.focus() });
 
 .notification-item__icon {
   flex-shrink: 0;
+  // A fixed footprint rather than the glyph's own. The icon comes from the
+  // notification type, its glyphs are not all the same width, and a column of
+  // titles that steps in and out by a few pixels a row reads as a list that
+  // cannot line itself up. 1.25em is the width a duotone glyph is drawn in, so
+  // nothing is clipped and the placeholders can reserve the same.
+  width: 1.25em;
   margin-top: 2px;
   color: $gray-lighter;
   font-size: 1.15em;
+  text-align: center;
 }
 
 .notification-item--unread .notification-item__icon {
--- a/app/frontend/admin/pages/notifications.vue
+++ b/app/frontend/admin/pages/notifications.vue
@@ -11,6 +11,7 @@ import Heading from "@/shared/components/base/Heading/index.vue";
 import HeadingSmall from "@/shared/components/base/Heading/Small/index.vue";
 import BtnGroup from "@/shared/components/base/BtnGroup/index.vue";
 import FilteredList from "@/shared/components/FilteredList/index.vue";
+import Empty from "@/shared/components/Empty/index.vue";
 import ListSkeleton from "@/shared/components/ListSkeleton/index.vue";
 import { useReportListGeometry } from "@/shared/composables/useListGeometry";
 import BulkSelectionBar from "@/shared/components/BulkSelectionBar/index.vue";
@@ -136,6 +137,10 @@ const list = ref<ComponentPublicInstance>();
 // is as tall as its own two lines or the controls beside them, whichever wins,
 // and only the rendered row knows which.
 useReportListGeometry("row", list, {
+  // Named rather than injected: the geometry is provided by the FilteredList
+  // this page renders, and inject only looks upwards - so the measurements of
+  // a list drawn in that list's slot have to be filed under its name.
+  name: "admin-notifications",
   ready: () => !!records.value.length,
   pick: (host) => host.firstElementChild,
 });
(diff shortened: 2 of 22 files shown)


mortik:
## Screenshots

Shot from the `/visual-tests/` routes, once at `main` (`eb412270d`) and once at this branch. The three demo pages this branch touches were checked out into the base worktree first, so both passes render the **same** harness and only the component code differs — otherwise the new demo sections would simply be absent from the "before" column.

### ListSkeleton beside the rows it stands in for — 1600px

Placeholder insets, measured from the column's left edge:

| | checkbox | icon | first line |
|---|---|---|---|
| record row | 35px | 96px | 130px |
| placeholder, before | — | 35px | 69px |
| placeholder, after | 34px | 93px | 128px |

| Before | After |
| --- | --- |
| Placeholder rows without a checkbox, icon and text bars sitting 61px left of where the record rows put them | Placeholder rows with a checkbox box, icon and text bars aligned with the record rows beside them |

Before, the icon sits where the checkbox belongs and both lines land 61px left of the title they stand in for.

### BulkSelectionBar in a phone's column — 1600px

| Before | After |
| --- | --- |
| Bulk selection bar wrapped to two lines with the select-all checkbox centred below the count and outside the column | Bulk selection bar wrapped to two lines with the select-all checkbox on the same line as the count |

Once the bar wraps, the checkbox used to centre itself against the whole two-line bar and left the count's line entirely. The separate `×` is gone too: the minus box now clears the hand-picked rows itself, which is what the demo copy describes.

### ListGroup holding itself open — 2200px

| Before | After |
| --- | --- |
| A loading list group collapsed to an empty box reading No Docks found, with the loader over the text | A loading list group holding four placeholder rows matching the four record rows beside it |

The pair is `col-xl-6`, so it only sits side by side above 2100px — at 1600px it stacks and the comparison the section is built on is not visible. The cube loader drawing over the placeholder rows is deliberate: `loaderVisible` only stands down where a surrounding frame is already showing one.

### Empty in a 361px column — 1280px

| Before | After |
| --- | --- |
| Empty box wider than its dashed 361px column, headline running off the right edge | Empty box inside its dashed 361px column, headline wrapped to two lines |

### Also checked, not posted

430×900 for the skeleton pair: the two columns stack, so the side-by-side check does not exist at that width — the insets hold. A clip that tall also carries the fixed bottom nav printed across its middle, which reads as an artifact rather than as the fix.

One thing the shots show that this branch does not change: in the demo the skeleton stands without a frame, so a record row whose title wraps to two lines is taller than the placeholder standing in for it and the two columns drift apart from that row down. On the real page the frame remembers the measurement (`geometry.heightFor("row")`), which is exactly the case that comment calls out.

🤖

PR #4851, opened 2026-09-10
feat(hangar): record what a user changes on their own ship

### LINKED ISSUE #4845: Record what a user changes on their own ship
Nothing a user does to their own ship is recorded. Split out from the change-history feature so the recording can be configured now — history cannot be backfilled, and every day this waits is a day of history that does not exist.

## Where it stands

`Vehicle` has paper_trail, but with its own crippled configuration rather than the shared one:

```ruby
has_paper_trail on: %i[update],
  only: %i[name serial wanted flagship public name_visible sale_notify hidden loaner bought_via],
  if: ->(record) { record.author_id.present? }
```

`author_id` is set only by an admin action. The result, measured against production: **zero `Vehicle` versions exist.** For comparison, `Fleet` has 553,744.

`vehicle_loadouts`, `vehicle_modules` and `vehicle_upgrades` have **no paper_trail at all** — so loadout changes, the most interesting thing a history would show, are not recorded anywhere.

## What is already in place

- `Vehicle` is registered in `app/lib/versioned_item.rb`, in both `ROOTS` and `POLICIES`. The admin read path exists.
- `Api::BaseController:145` sets `whodunnit` from `current_resource_owner`, so a user's edits get an actor for free the moment they are recorded at all.
- `Inventory`, `InventoryItem`, `FleetInventory` and `FleetInventoryItem` already use the shared `has_paper_trail on: ::VersionedItem::RECORDED_EVENTS`, unconditionally. That is the pattern to adopt.

## The real work: keeping machine writes out

The `author_id` condition is doing a job, and removing it without a replacement would flood the table. Vehicles are written automatically in several places:

- `create_loaner` and `create_bundled_snub_craft` create rows on every parent save
- a hangar sync rewrites many vehicles at once
- today's loaner repair touched roughly 54,000 rows

`VersionedItem::RECORDED_EVENTS` documents what this looks like when it goes wrong: `Fleet` carries 572,735 empty touch versions against 464 real edits. The same trap, from a different direction.

So this issue is a design decision, not a config change: what counts as a change to a ship, and how machine writes are held out — a `loaner? || bundled?` guard, `PaperTrail.request(enabled: false)` around the importers, or something else.

## Open decision

**Do loadout, module and upgrade changes get recorded too?** They have no instrumentation today and they are what makes a history worth reading. Including them adds three models to this issue; leaving them out means the later history tab is missing its most interesting column.

## Not in scope

The user-facing tab. The existing read path is admin-only and policy-gated, and it is worth seeing what the recorded data actually looks like before designing a view onto it.

🤖

mortik:
Resolves #4845.

## The problem

Nothing a user does to their own ship is recorded. `Vehicle` has had paper_trail all along, gated on `author_id.present?` — and `author_id` is not a column. It is an `attr_accessor` (`vehicle.rb:47`) assigned in exactly two places, both admin: `Admin::Api::V1::VehiclesController#update` and `Versions::FieldReverter`. Nothing a user does can satisfy it.

Measured against production:

```
Fleet             553744
FleetMembership    46745
Model               8503
RoadmapItem         2523
Component            480
Vehicle                0
```

Zero, against a table of **1,573,370 rows**. History cannot be backfilled, so every day this waits is a day that does not exist.

## The shape

The issue framed this as a design decision rather than a config change, and the reading confirmed it: **no predicate on the record can tell a machine write from a user's.** `HangarSync#sync_vehicles` calls `vehicle.update!` at ~8 sites writing `name` and `wanted` — the same columns, on the same non-loaner rows, as a rename.

Both candidate guards leak:

| Guard | Holds out | Leaks |
|---|---|---|
| `loaner? \|\| bundled?` | the loaner and bundled cascades | `HangarSync`, `LoanerJob`, `Rsi::ModelsLoader`, `HangarImporter` |
| `whodunnit.present?` | the three Sidekiq paths | `HangarImporter` — it runs inline in the request, so the user *is* the whodunnit |

`loaner? || bundled?` holds out 604,464 of 1,573,370 rows and leaves 968,906 user-owned rows exposed to sync. So the two jobs are split: the predicate keeps out the rows that have no history worth reading, and the importers are silenced at their entry points with `PaperTrail.request(enabled: false)` — a new mechanism here, it appeared nowhere in the codebase before.

Held out at the entry point rather than at each write site, so a write added to a sync later is silent by default. The failure mode is a missing version, not a flood. In the models loader only the vehicle writes are wrapped — `model.destroy` still files its own `Model` version.

`only:` also gains `model_id`, `model_paint_id` and `alternative_names`. All three were user-writable but unrecorded, and a repaint and a retrofit are exactly what a history should show.

## Two of the issue's premises were wrong

**The loaner repair was never a flood risk.** `Maintenance::RepairLoanerFlagsTask` writes through `update_columns`, which bypasses paper_trail entirely. The ~54,000 rows it touched would have filed nothing, so it is not one of the paths that needed holding out.

**`:destroy` cannot be recorded on `Vehicle`,** so adopting `VersionedItem::RECORDED_EVENTS` verbatim as the issue suggested would have added an event that provably writes nothing. `ErasableVersionsConcern` is included before `has_paper_trail` and rails runs `after_*` in reverse definition order, so the destroy version is written and then deleted in the same transaction. An existing test already pins that outcome.

## The open decision: loadouts yes, modules and upgrades no

The answer is not symmetrical, which is why it took reading the write paths rather than the models.

`VehicleLoadout` is written only by its own controller, one `save` per user action. One action, one version. It records cleanly.

`VehicleModule` and `VehicleUpgrade` are torn down and rebuilt on every update:

```ruby
@vehicle.vehicle_modules.destroy_all unless vehicle_params[:model_module_ids].nil?
@vehicle.vehicle_upgrades.destroy_all unless vehicle_params[:model_upgrade_ids].nil?
```

Any PATCH that merely *includes* `modelModuleIds` would file a destroy and a create for every module on the ship, changed or not. That is churn, not change. Recording them meaningfully needs the controller to diff instead of rebuild, which is a separate change.

## One thing found while implementing, not in the issue

`Vehicle has_many :vehicle_loadouts, dependent: :destroy`, and `User has_many :vehicles, dependent: :destroy`. A loadout version carries `name` — free text the owner typed — and `url`, their build. Without `ErasableVersionsConcern` those rows outlive a deleted account, which is precisely what that concern exists to prevent on `Vehicle`.

So `VehicleLoadout` includes it. That costs the delete event for the same reason as above — the concern erases the row paper_trail just wrote — so removing a loadout is not recorded. That is the trade a ship already makes for being deleted, and consistency with that decision beat keeping one more event; the alternative erases a loadout's versions when the *vehicle* dies but not when the loadout does, which needs ids captured before destroy.

## Verification

98 tests, 315 assertions, 0 failures across the model, loader, job and admin-versions suites. `bin/ruby-lint` clean over 2,514 files.

The two wrappers are pinned end to end rather than only in the abstract: a real `HangarSync` over three existing vehicles files zero versions while still renaming the corsair and wishlisting the javelin, and a real `HangarImporter` run files zero while still building the hangar. Both would have flooded before.

## Notes

- `VehicleLoadout` joins `VersionedItem::ROOTS` as `[:vehicle]` so the admin read path can reach it, authorised through the ship. `TYPES` feeds an OpenAPI enum, hence the one-line schema regeneration.
- It stays out of the recent feed, which filters `.where.not(author_id: nil)` and would show nothing for it.
- **User edits will render as `author: null` in the admin history.** The views resolve the author from `AdminUser` via `author_id` and never render `whodunnit`, so the data is attributed but the view cannot read it. That is a read-path change and the issue scopes the read path out — worth knowing before anyone reads the recorded data as broken.
- `move_all_ingame_to_wishlist` keeps recording. It is a genuine user change to every ship it touches and bounded by their own hangar, but it is the one user path that can file hundreds of versions from one press — the first place to look if `versions` grows faster than expected.
- The `versions` table has two indexes and the `created_at` one is partial on `author_id IS NOT NULL`, so user versions fall outside it. Nothing queries user history by time until the tab exists.
- `VehicleLoadout#activate!` uses `update_all`, so activating B records B but not that A stopped being active. Asymmetric, left alone.
- The user-facing history tab stays out of scope, per the issue — worth seeing the recorded shape first.

🤖

## Summary by CodeRabbit

* **New Features**
  * Vehicle changes made by owners now appear in version history, including model, paint, and alternative-name updates.
  * Vehicle loadout creation and edits are now tracked, with changes attributed to the responsible user.
  * Vehicle history now supports loadout-related entries in the admin API.

* **Improvements**
  * Automated imports, synchronization, loaner updates, and cleanup operations no longer create version-history entries.
  * Admin vehicle edits continue to record administrative attribution separately from owner changes.

### DIFF
--- a/app/jobs/loaders/loaner_job.rb
+++ b/app/jobs/loaders/loaner_job.rb
@@ -7,13 +7,17 @@ module Loaders
 
       model_ids = ModelLoaner.pluck(:model_id).uniq
 
-      model_ids.each do |model_id|
-        Vehicle.where(model_id:, loaner: false).find_each(&:add_loaners)
+      # `add_loaners` runs over every vehicle of every loaner-bearing model, so
+      # this is the widest machine write there is against the hangar.
+      PaperTrail.request(enabled: false) do
+        model_ids.each do |model_id|
+          Vehicle.where(model_id:, loaner: false).find_each(&:add_loaners)
+        end
+
+        loaner_model_ids = ModelLoaner.pluck(:loaner_model_id).uniq
+
+        Vehicle.where(loaner: true).where.not(model_id: loaner_model_ids).destroy_all
       end
-
-      loaner_model_ids = ModelLoaner.pluck(:loaner_model_id).uniq
-
-      Vehicle.where(loaner: true).where.not(model_id: loaner_model_ids).destroy_all
 
       actionable = missing_loaners.present? || missing_models.present?
 
--- a/app/lib/hangar_importer.rb
+++ b/app/lib/hangar_importer.rb
@@ -29,61 +29,65 @@ class HangarImporter
     missing_models = []
     imported_models = []
 
-    (@import.import_data || []).each do |item|
-      name = item[:name]
-      name = legacy_mapping[item[:name]] if legacy_mapping[item[:name]].present?
-      name = starship_42_mapping[item[:name]] if starship_42_mapping[item[:name]].present?
-      name = hangar_xplor_mapping[item[:name]] if hangar_xplor_mapping[item[:name]].present?
+    # An import creates the user's hangar wholesale; none of it is an edit the
+    # user made to a ship that already existed.
+    PaperTrail.request(enabled: false) do
+      (@import.import_data || []).each do |item|
+        name = item[:name]
+        name = legacy_mapping[item[:name]] if legacy_mapping[item[:name]].present?
+        name = starship_42_mapping[item[:name]] if starship_42_mapping[item[:name]].present?
+        name = hangar_xplor_mapping[item[:name]] if hangar_xplor_mapping[item[:name]].present?
 
-      normalized_name = normalize(name)
-      slug = item[:slug].downcase if item[:slug].present?
-      slug = item[:paint_slug].downcase if item[:paint_slug].present?
+        normalized_name = normalize(name)
+        slug = item[:slug].downcase if item[:slug].present?
+        slug = item[:paint_slug].downcase if item[:paint_slug].present?
 
-      query = [
-        MODEL_FIND_QUERY.join(" OR "),
-        {
-          name: name.downcase,
-          slug:,
-          normalized_name:,
-          search: "%#{normalized_name}%"
+        query = [
+          MODEL_FIND_QUERY.join(" OR "),
+          {
+            name: name.downcase,
+            slug:,
+            normalized_name:,
+            search: "%#{normalized_name}%"
+          }
+        ]
+
+        params = {
+          notify: false,
+          user_id: @import.user_id,
+          name: item[:ship_name] || item[:custom_name],
+          serial: item[:ship_serial],
+          flagship: item[:flagship] || false,
+          wanted: item[:wanted] || !item[:purchased] || true,
+          bought_via: item[:bought_via] || :pledge_store,
+          public: item[:public] || false,
+          name_visible: item[:name_visible] || false,
+          sale_notify: item[:sale_notify] || false,
+          hangar_group_ids: HangarGroup.where(user_id: @import.user_id, name: item[:groups]).pluck(:id),
+          model_module_ids: ModelModule.where(name: (item[:modules] || []) + (legacy_module_mapping[item["name"]] || [])).pluck(:id),
+          model_upgrade_ids: ModelUpgrade.where(name: item[:upgrades]).pluck(:id)
         }
-      ]
 
-      params = {
-        notify: false,
-        user_id: @import.user_id,
-        name: item[:ship_name] || item[:custom_name],
-        serial: item[:ship_serial],
-        flagship: item[:flagship] || false,
-        wanted: item[:wanted] || !item[:purchased] || true,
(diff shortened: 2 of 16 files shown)


mortik:
Good catch, and fixed in 7d3ee85 — this was a gap the plan had listed as deferred, on the assumption that the fix traded one bulk write for one write per sibling.

That assumption was wrong, and measuring it flipped the decision. The old call wrote `active: false` to *every* non-self sibling, nearly all of them already false. Scoping to `.active` and iterating writes only the row that was actually on — and there is at most one, because that is the invariant `activate!` itself maintains. Production agrees: 273 loadouts across 221 vehicles with an active one, max 1 active per vehicle, 0 violations, and only 5 vehicles hold more than one loadout at all.

So it records strictly more history while doing strictly fewer writes. Nothing to defer.

Two tests cover it: one asserts the changeset on both sides of the switch (`[false, true]` on the arriving loadout, `[true, false]` on the one it replaced), and one asserts an already-inactive sibling files nothing — which is what would regress if this ever went back to touching every row.

🤖

PR #4874, opened 2026-09-11
fix(sc-data): read the Nox with its length on z

mortik:
Left on the default axis order the Nox reads **1.45 m long and 5.15 m tall** — a hover bike stood on its end.

Both dimension columns agreed on it, because `length/beam/height` and `sc_*` trace back to the same bounding box. So nothing flagged it: it is not a drift between curated and loaded values, it is the order being wrong at the source for both.

`%i[z x y]` — the same permutation as the Caterpillar, from the same source, the proportions read off the orthographic renders. That gives 5.15 × 1.32 × 1.45, which is a bike.

Covers the Kue, which is the same chassis.

## Note

This takes effect on the **next model load**. The stored rows keep the old order until then, so the Nox stays wrong in production until a load runs.

## Verification

- 11 loader tests green, one of them new and shaped like the existing Caterpillar and Cyclone cases
- `ruby-lint` clean

🤖

### DIFF
--- a/app/lib/sc_data/loader/models_loader.rb
+++ b/app/lib/sc_data/loader/models_loader.rb
@@ -34,6 +34,8 @@ module ScData
         "rsi_scorpius_antares" => %i[x y z],
         "rsi_hermes" => %i[x y z],
         "tmbl_cyclone" => %i[x y z],
+        "xian_nox" => %i[z x y],
+        "xian_nox_kue" => %i[z x y],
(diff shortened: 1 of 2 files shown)


mortik:
Guter Fang — der Kue war eine eigene Zeile in `AXIS_ORDER` und wurde von keiner Zusicherung berührt. Der Test läuft jetzt über beide Schlüssel.

🤖

PR #4886, opened 2026-09-13
feat(hangar): let a sync decide what happens to ships it cannot find

### LINKED ISSUE #4884: Let a sync decide what happens to ships it does not find
Follow-up to #4879 / #4880.

A hangar sync moves every pledge-store ship it does not find in the RSI list onto the wishlist. That is one hard-coded answer to a question that has several reasonable ones, and it is the destructive one when a scrape comes back partial.

The run should carry the choice:

- **Move to the wishlist** — today's behaviour, and the default.
- **Delete them** — for people who melt and never look back.
- **Leave them unchanged** — the safety valve when a sync returns a short list.
- **File them into a hangar group** — left owned and untouched, just collected somewhere so they can be sorted out by hand.

Remembered between runs, the same way `addBundledVehicles` is.

🤖

mortik:
Closes #4884. Follow-up to #4880.

## The problem

A sync moved every pledge-store ship it could not match against the RSI list onto the wishlist:

```ruby
vehicle_scope.where.not(id: vehicle_ids).find_each do |vehicle|
  vehicle.update!(rsi_pledge_id: nil, rsi_pledge_synced_at: nil, wanted: true)
end
```

One hard-coded answer to a question with several reasonable ones — and the destructive one when a scrape comes back short, because a half-loaded RSI page moves the untouched half of a hangar onto the wishlist.

## The change

The run carries the choice, remembered between syncs the way `addBundledVehicles` is.

| value | what happens to an unmatched vehicle |
| --- | --- |
| `wishlist` | today's behaviour, and the default |
| `delete` | `Vehicle.delete_with_dependents` — loaners, bundled crafts, loadouts and versions go too |
| `keep` | nothing is written at all |
| `group` | nothing is written to the ship; it is filed into a hangar group to be sorted out by hand |

`group` gets a group of its own. `hangar_group_id` already means "file the ships it *did* match here", and one group for both would tip the matched and the unmatched into the same bucket.

## Two things the option surfaced rather than introduced

**The unmatched set is now scoped to `purchased`.** A wishlisted ship is one the user does not own, so it is absent from every RSI hangar and unmatched on every single run — `delete` would have emptied the whole wishlist on the first sync. The wishlist move never touched those rows either: `reset_pledge_id_if_wanted` has already cleared what it writes, so the `update!` was a no-op the `updated_at` check kept out of the report. Behaviour-preserving for the default, load-bearing for the rest.

**`keep` reports what it left alone.** Without that the option is a silent no-op and the user never learns which ships the sync failed to find.

The `@cancelled` guard covers all four. That already mattered for the wishlist move; under `delete` an unreached vehicle is not recoverable.

## Surface

- `imports.unmatched_vehicles_action` (string, default `wishlist`, not null) and `imports.unmatched_hangar_group_id`
- `PUT /hangar/sync-rsi-hangar` takes `unmatchedVehiclesAction` + `unmatchedHangarGroupId`. An absent or unrecognised action falls back to `wishlist`; `group` without a group the user owns falls back to `keep` rather than 400ing away the whole scrape
- `output` gains `deletedVehicles` (names — nothing is left to resolve an id from), `groupedVehicles`, `unchangedVehicles`

## Modal

Two smaller fixes the new row forced:

- The intro paragraph claimed unmatched ships "will be moved to your Wishlist". That stopped being true the moment the choice existed, so the clause is gone — the select below it answers that question now.
- The existing group picker said "Group", which was fine while it was the only one. It now says which ships it files, so the two pickers cannot be read for each other.

## Tests

- Loader: one per action; loaner/group cleanup on delete; a wishlisted ship never reached; a cancelled run deleting nothing; a re-sync not stacking task forces
- Endpoint: the choice is recorded, an absent action still moves to the wishlist, a group belonging to somebody else falls back to `keep`
- Vitest: the default, the chosen action, the group sent only under `group`, and start disabled until a group is picked
- Playwright: the default reads back, the group picker is conditional, the choice outlives a reload

Rebased onto main after #4880 was squashed in.

🤖

## Summary by CodeRabbit

* **New Features**
  * Hangar synchronization now lets you choose how unmatched vehicles are handled: move to wishlist, delete, keep unchanged, or assign to a group.
  * Group-based handling includes a target group selector.
  * Sync results and import details now show deleted, grouped, and unchanged vehicles.
  * Sync preferences are saved for future runs.
* **Translations**
  * Added localized labels and notifications for the new settings and outcomes.
* **Tests**
  * Added coverage for action selection, group assignment, persistence, and result reporting.


mortik:
@greptileai review

🤖

mortik:
Both findings were real, and both were gaps the `delete` action turned from survivable into not. Fixed in 59be22d.

**Stale cancellation** — correct. `@cancelled` is only set at the ship loop's checkpoints, so a cancel landing after the last one leaves it false while the row already says cancelled. The row is now read once more immediately before reconciliation. Covered by a test that cancels from the call directly preceding it, which is the one window no checkpoint covers; it fails without the fix.

**Deleted groups invalidating syncs** — also correct, and worse than described in one respect: it raises mid-run, not only at `start!`. Verified by restoring the validation, at which point the test dies at `import.update!(output:)` with `Validation failed: Unmatched hangar group can't be blank` — after the run has already done its writes.

Rather than keeping the validation and working around it, it is gone. It was the wrong place for the constraint: the request boundary already normalises the pairing (a `group` action with no group the user owns falls back to `keep`), and a group that disappears mid-flight is a legitimate state the run has to survive, not an invalid record. A vanished group now leaves the ships alone and reports them as `unchangedVehicles`, rather than claiming they were filed somewhere they are not.

🤖

mortik:
@greptileai review

Both P1s are fixed in 59be22d — the summary above still names c667811, which predates them.

🤖

PR #4889, opened 2026-09-13
feat(relationships): friends for users, allies for fleets

### LINKED ISSUE #4887: Friends for users, allies for fleets
## Description

Follow-up to #4878 / #4881. Transfers gave a user and a fleet a *stance* on who may send them goods — `everyone`, `known`, `nobody` — and `known` currently has exactly one meaning: you share a fleet with me. That is the only relationship the platform can express. There is no way to say "this person is a friend" or "this org is our ally", so the two things that relationship should unlock are unavailable:

- a hangar or wishlist that is not public, but is open to the people you actually play with;
- a fleet that does not publish its ships, but shares them with the org it flies alongside.

Two relationships, both symmetric, both agreed to rather than declared:

| | between | managed by |
|---|---|---|
| **Friendship** | user ↔ user | each user, for themselves |
| **Alliance** | fleet ↔ fleet | each fleet's admins |

Both use the same two-step handshake: one side **sends a request**, the other side **accepts, declines, or ignores** it. Ignoring is not a slower decline — the sender must not be able to tell it happened, otherwise it is a decline with extra steps.

What the relationship then unlocks:

**Friends** — when the owner has switched it on, a friend can see a hangar and a wishlist that are not public, and the hangar stats that go with them. They are also a transfer party: a friend satisfies the `known` policy and appears in the transfer modal's party picker without having to share a fleet.

**Allies** — when the fleet has switched it on, an allied fleet's members can see the fleet's ships, its stats, and its member list. An allied fleet satisfies `known` on both ends and appears as a target when sending on a fleet's behalf, so fleet-to-fleet logistics works between orgs that are not the same org.

Neither relationship grants anything by itself. Every surface stays behind a per-surface setting the owner controls, and a relationship only ever widens an audience from "nobody" to "my friends" / "our allies" — never past a setting that is already off, and never past an existing transfer denial.

## Acceptance Criteria

- [ ] A user can send a friend request to another user by username, and cancel one they sent
- [ ] The recipient can accept, decline, or **ignore** it; ignoring is indistinguishable from an unanswered request on the sender's side, and absorbs any further request from that sender
- [ ] Either friend can end the friendship unilaterally
- [ ] A fleet's admins can send, cancel, accept, decline and ignore alliance requests to and from another fleet, with the same three answers
- [ ] New per-surface visibility settings let a user open their hangar, wishlist and hangar stats to friends without making them public
- [ ] New per-surface settings let a fleet open its ships, its stats and its member list to allied fleets without making them public
- [ ] An ally's view of the member list is a reduced payload — it must not carry the contact handles, coordinates, current system or activity timestamps the internal member list does
- [ ] A member who has hidden themselves is still hidden from an allied fleet
- [ ] A friend satisfies the `known` inventory transfer policy for user↔user; an alliance satisfies it for fleet↔fleet
- [ ] Friends appear in the transfer party picker; allied fleets appear as targets when sending on a fleet's behalf
- [ ] A standing transfer denial still refuses a friend or an ally — a relationship widens a stance, it does not override an exception to one
- [ ] Both request flows produce notifications for the request and for its acceptance, and none for a decline or an ignore
- [ ] Both surfaces are behind feature flags and ship without changing anything for anyone who does not use them

## Notes

- Depends on #4881. The transfer half plugs into `Inventories::TransferGate#shares_a_fleet?` and the `useTransferTargets` composable, whose comment already names a friends list as its intended extension point. Work happens on `feat/4887-friends-and-fleet-allies`, stacked on `feat/4878-inventory-transfers`.
- Plan: [`docs/exec-plans/4887-friends-and-fleet-allies.md`](https://github.com/fleetyards/fleetyards/blob/feat/4887-friends-and-fleet-allies/docs/exec-plans/4887-friends-and-fleet-allies.md)

🤖

mortik:
Transfers gave every party a stance on who may send it goods, and `known` had exactly one meaning: you share a fleet with me. `TransferGate#shares_a_fleet?` was the whole of it, and it was the only relationship this platform could express. Everything else was a binary — published to the world, or closed.

This adds the two relationships that were missing, and the things they unlock.

Resolves #4887 · Plan: [`docs/exec-plans/4887-friends-and-fleet-allies.md`](https://github.com/fleetyards/fleetyards/blob/feat/4887-friends-and-fleet-allies/docs/exec-plans/4887-friends-and-fleet-allies.md)

| | between | managed by |
|---|---|---|
| **Friendship** | user ↔ user | each user, for themselves |
| **Alliance** | fleet ↔ fleet | each fleet's admins |

## The four decisions worth reviewing

**Ignoring is invisible, and that is the whole feature.** An ignore the sender can detect is a decline with worse manners, so `ignored` renders to the sender exactly as `pending` does — and a further request from them is *absorbed*: the API answers as though it were sent, the row does not move, nothing reaches the recipient. That absorption is why this ships with no block list; a permanent, silent, per-relationship refusal is what blocking would have been for here, and it costs one state instead of a table.

Three things gave it away during the build and each is fixed on purpose. A list filtered on the column omitted the row from the sender's own pending list, so listing goes through `in_state_for(state, party)` rather than ransack. `cancellable_by?` refused to withdraw an ignored request, so it is asked of the state the sender can *see*, and cancelling one answers success and leaves the row standing — destroying it would make withdraw-and-ask-again a way back into an inbox that turned you away. And nothing is notified: a request absorbed by an ignore reaches nobody, which is its own test.

**Visibility is a second boolean per surface, not an enum.** `public_hangar` and its four siblings are booleans in the published schema and on `/users/me`; turning them into three-valued enums is a breaking change. So each surface gains a sibling and the rule is stated once: **public wins, and the narrower switch is only consulted when it is off**. It is a three-valued enum spelled in two booleans, and that is the honest description of the trade — what it buys is that no existing value changes meaning. Consolidating them belongs in a deliberate breaking release, all five at once.

Three call sites did not go through `Public::UserPolicy` and each would have kept the old answer: the deprecated fleetchart action checked the column inline *twice*, the multi-hangar embed filtered a whole list of usernames on it, and the stats page decided its wishlist total the same way. The embed gets a scope rather than a predicate because it asks about a list at once, and a test asserts scope and policy agree on all eight combinations.

**An ally's roster is a different payload, not the same one behind a policy.** `fleet_members/_base.jbuilder` renders each member's Discord, YouTube, Twitch and Guilded handles, their RSI handle, their homepage, their **latitude and longitude**, their current system and when they were last active. Handing that to another organisation because two admins shook hands is not "showing the member list", and a policy in front of it would hand over every field anybody ever adds to it. So: a separate controller and a four-field partial, asserted over the whole key set. A member with `hide_owner` keeps their place and loses their name, as they already do on a vehicle they own.

Ship visibility needed no new rule at all — `fleet.vehicles` reads `FleetVehicle` rows each member's own `ships_filter` already decided, so an ally inherits every member's choice for free. There is a test that it stays that way.

**`known` gained exactly two meanings, and nothing else about transfers changed.** A friendship for user↔user, an alliance for fleet↔fleet. It does not reach `check_rule`, so a standing denial still refuses a friend; it does not soften `nobody`. Both have tests. On the client, `useTransferTargets` grows the entry its own comment predicted — friends become another group to pick a person out of — and allied fleets become targets only when sending on a fleet's behalf, since an alliance is between the two organisations and not between one of them and each member of the other.

## Worth knowing

- **Alliances default to `fleet:manage`.** New `fleet:allies:*` privilege group; officers read, members see nothing. Committing the fleet's ships, stats and roster to another org is not an officer's call. The client gates on evaluated capabilities, never on privilege strings.
- **Two flags, `friends` and `fleet_allies`**, because `FeatureSetting` carries the user and fleet self-service scopes independently — the same split `hangar_inventories` and `fleet_logistics` already make.
- **Two notification types per relationship, not four.** No decline notification and obviously no ignore one. The membership types have never had a declined counterpart either, for the same reason. Friend requests can be answered from the notification itself; an alliance notification links to the allies page rather than guessing which of the two fleets the reader acts for.
- **85 new `oasdiff-ignore` lines**, from eight enum values landing on published responses — the file already carries 907 of these. 80 match an existing line except for the value; the five for `record/type` are the first notification *record* type added since the file was written, so they have no counterpart to check the wording against. Written with oasdiff 1.31.0 against CI's pinned 1.18.1 — the rule and message are the same in both and the property path comes from the schema, but that job is the one to glance at.
- **Rebased onto main** after #4881 merged mid-build. Two conflicts, both resolved keeping both sides.
- **One plan item not delivered:** the Playwright pass over request → accept → the newly visible hangar. Everything else is covered by Minitest and Vitest.

🤖


mortik:
@greptile-apps please review.

Most of the 177 files are generated or mechanical — `swagger/`, `asyncapi/`, the oasdiff ignore file, and 42 locale files. The parts worth your attention:

**The invariant the whole feature turns on** — an ignored request must be indistinguishable from an unanswered one, *to the sender*, on every surface:
- `app/models/concerns/party_relationship.rb` — `state_for`, `in_state_for`, `cancellable_by?`
- `app/services/relationships/requester.rb` — the four cases a new request can meet, and why the capacity check runs *after* the ignored branch
- `app/services/relationships/answerer.rb` — cancelling an ignored request answers success and changes nothing
- `app/services/relationships/notifier.rb` — an absorbed request notifies nobody

**Visibility**, where a missed call site is a leak rather than a bug:
- `app/policies/public/user_policy.rb`, `app/policies/public/fleet_policy.rb`
- `app/controllers/api/v1/public/hangars_controller.rb` (`#embed` — the bulk form of the policy) and `app/models/user.rb` (`with_hangar_readable_by`)
- `app/controllers/api/v1/public/fleet_members_controller.rb` + `app/views/api/v1/public/fleet_members/_base.jbuilder` — deliberately not the fleet's own member payload

**Authorization**: `app/policies/friendship_policy.rb`, `app/policies/fleet_alliance_policy.rb`, `app/controllers/concerns/relationship_actions.rb`.

**The one line that changes transfers**: `known_to_each_other?` in `app/services/inventories/transfer_gate.rb`.

Two things I would especially like a second pair of eyes on:

1. Whether any path lets a sender tell an ignored request from a pending one — response body, status code, list membership, or notification.
2. Whether an allied fleet can reach anything beyond ships, stats and the four-field roster.

🤖

mortik:
@greptile-apps please re-review — all four findings are addressed in `ba864a2`. Three were right and one of them was right about something my own plan had already warned against.

| finding | fix |
|---|---|
| Withdrawal reveals ignored requests | `withdrawn_at` on both tables |
| Create permission bypasses manage | `accept?` authorized on the crossing path |
| Stats sharing exposes fleet | `allies_fleet_stats` no longer satisfies `show?` |
| Request transitions can race | existing-row branches under a row lock |

**Withdrawal.** Correct, and the sharpest of the four — the ignore was invisible everywhere except in what happened *after* withdrawing one. A real withdrawal destroys the row; an ignored one had to survive to keep absorbing, so it came back on the next fetch. It is now marked `withdrawn_at` rather than left untouched: excluded from `in_state_for`, 404 on a direct read (`set_relationship`), still absorbing, and cleared by a fresh request so the row reads pending again. `friendship_actions_test.rb` now runs withdraw → read → list → re-request against an ignored row and a plain one and asserts the two answer identically.

**Create vs manage.** Correct. `RelationshipActions#create` now looks for the inbound pending row before calling the service and authorizes `accept?` on it when there is one. No change for friendships, where both rules are "you"; for a fleet, a `fleet:allies:create` role gets 403 on the crossing path and still 201 on a request of its own. Both are tested.

**Stats sharing.** Correct, and worth spelling out because I had written the opposite into the plan and then not followed it. D7 says "the ally clause has to be added to each rule rather than to a shared helper: folding them together would let `allies_fleet_stats` quietly open the fleet page too" — and I added it to `show?` anyway, reasoning from `public_fleet_stats?` already being there. What that missed is that `show?` is not just the profile: `Public::FleetVehiclesController` authorizes against it, so the switch labelled "Stats visible to allies" was also handing over the ship list with loadouts, modules, upgrades, groups and owner avatars. Now one switch, one surface. I left the pre-existing `public_fleet_stats?` clause alone — same shape, but changing it would alter published behaviour and belongs in its own change.

**Race.** Correct, low impact, cheap to close. The existing-row branches re-read under `lock` inside a transaction, the way `Answerer` already does. Notifications moved *outside* that transaction while I was there: a notification writes a row and broadcasts, and neither should be able to roll back a relationship that was legitimately agreed to.

Two things I would still value a second look at, since they are the invariants the rest hangs on:

1. Any remaining path where a sender can tell an ignored request from a pending one — including the new withdrawn state.
2. Whether an allied fleet can now reach anything beyond ships, stats and the four-field roster, each behind its own switch.

🤖

mortik:
@greptile-apps please take one more pass — your 5/5 was on `ba864a2` and two commits have landed since, one of which touches code you reviewed.

`205e135f` fixes three things CI caught, none of them behavioural except the last line of each error response:

- **Validation error codes.** `RelationshipActions` rendered a bare `"create"`/`"destroy"` and an *interpolated* `event.to_s` for the three answers. `ValidationErrorTest` scans `app/` for literal `ValidationError.new("...")` codes and checks each resolves in all seven locales — an interpolated one is invisible to that scan, and a missing key does not raise, it renders `"translation missing: ..."` into the payload. All five are literal and namespaced now (`relationships.accept`, …), which is why `accept`/`decline`/`ignore` are three one-line methods rather than one generated helper.
- **`asyncapi/` was stale.** The four notification types and the `friendship` record type are enums a cable channel broadcasts, and `bin/generate-schema` does not touch that tree.
- **`useNotificationRecordActions.spec.ts`** mocks the fyApi hooks by name, so adding `useFriendship` sent all eight of its tests into the real vue-query hook. Mocked, plus tests for the friend-request branch.

`08150a8f` is the exec-plan only.

The invariants are unchanged, so the two questions from my last comment still stand — particularly whether anything now distinguishes an ignored request from a pending one for its sender, including the new withdrawn state.

🤖

PR #4892, opened 2026-09-13
fix(asyncapi): regenerate the cable schema for the landed fleetchart offsets

mortik:
`api-schema-check` has been red on main since af1a3bd0c ("feat(models): give the landed state its own fleetchart offsets"). That commit added `landed_fleetchart_offset_length` and `landed_fleetchart_offset_beam` to the model cable broadcast but did not regenerate the checked-in document, and the job is exactly `bin/generate-asyncapi` followed by `git diff --exit-code ./asyncapi/**/schema.yaml`.

The diff is the four lines the generator wants:

```yaml
         extendedFleetchartOffsetBeam:
           type: number
+        landedFleetchartOffsetLength:
+          type: number
+        landedFleetchartOffsetBeam:
+          type: number
         height:
           type: number
```

Generated output only — no hand edits. The local run produces the identical blob CI did (`077ffbf..fc3d5ff` in both).

Found while opening #4891, which inherits the red check from its base.

🤖


mortik:
Superseded by #4893, which carries this commit unchanged alongside the e2e fix.

The two had to be combined: this PR's only red check was `e2e-test (2, 2)`, which #4893's commit fixes, and #4893's only red check was `api-schema-check`, which this commit fixes. Split across two branches neither could ever go green, so neither could merge.

Branch left in place; the cherry-pick is byte-identical. 🤖

mortik:
Reopening — the squash merge of #4893 took only its e2e commit, so the schema regeneration never reached main and `api-schema-check` is still red there:

```
$ git show origin/main:asyncapi/cable/v1/schema.yaml | grep landedFleetchartOffset
(no match)
```

My "superseded" note above was wrong. The deadlock that prompted it is gone, though: main now has the e2e fix, and this branch has been updated with it, so this PR should go green on both checks on its own. 🤖

PR #4915, opened 2026-09-14
feat(supporters): link a donation to the account that made it

mortik:
Supporters can now be matched to the account that paid — and the plan for what that unlocks later.

Resolves nothing yet on its own; this is the first slice of [`docs/exec-plans/premium-fleet-features.md`](https://github.com/fleetyards/fleetyards/blob/feat/premium-fleet-features/docs/exec-plans/premium-fleet-features.md) (Phases 1–3), which is deliberately useful whether or not premium features ever ship.

## The problem, measured

`User#supporter?` has existed for a while under a comment reading *"Gate perks on this one"*. Nothing gates on it, and nothing could:

| | |
|---|---|
| Contributions active this month | 3, all Patreon |
| ...with a `user_id` | **0** |
| All contributions ever, with a `user_id` | **0 of 16** |

`SupporterImporter#assign_defaults` sets source, anonymity, recurrence and a start date. It never sets `user_id`, and the admin link field has never been used. So `supporter?` is `false` for every account in production, including the three that pay.

## Three ways in, one resolver

Every funding platform carries the payer's address, so `Supporters::Linker` takes a contribution rather than a platform, and `supporter_contributions.payer_email` is written by all of them — Patreon through its API, Ko-fi through its webhook, PayPal and Buy Me a Coffee through an admin typing it in.

Only **confirmed** accounts match. Neither side is a claim: the platform verified the address for billing and we verified it at confirmation, which is what makes this safe where a user-entered address would not be.

The email is stored rather than resolved-and-discarded, which buys the case that otherwise loses people silently — a donor whose email had no account when they paid is found when they register. The linker is idempotent and re-runs over unlinked rows.

**The claim key** covers the rest: a supporter whose platform email differs from their Fleetyards one. Crockford base32 in an `FY-XXXX-XXXX` shape — no `I`, `L`, `O` or `U`, and decoding folds the characters people substitute anyway, so a key transcribed slightly wrong still resolves.

## Decisions worth reviewing

**There is no key on Patreon, and there cannot be.** Patreon API v2 exposes no patron-written field — `note` is documented as *"The creator's notes on the member"*, and there is no pledge message or survey response on the Member resource. So the key is for the platforms that carry a donor message, and Patreon links by verified email instead. This is why identity splits by platform rather than picking one mechanism.

**A Ko-fi subscription payment is recorded as non-recurring.** Ko-fi notifies on payment and *never* on cancellation, so a `recurring` row would have no event that could ever end it and would grant perks forever. One row per month means `active_in` counts it only in the month it was paid — lapsing is automatic and needs no second signal.

**The linker runs outside the `changed?` guard** in the Patreon importer. A row identical to last week's still wants linking when the person behind it registered since. Asserted directly: `updated: 0, linked: 1`.

**An unverifiable Ko-fi request answers 401, not 400** — including a body that is not JSON. It cannot be verified, and an unauthenticated endpoint gains nothing by telling a caller which half of the request was wrong.

**Anonymity hides a name, not the badge.** The badge asked `public_supporter?`, which required `anonymous: false`, so an anonymous contribution earned perks and no badge. Anonymity decides whether a contribution is *named* on the supporters page — `public_name` already answers that, independently. `public_supporter?` had no other caller and is removed rather than left as a method whose name invites the old rule back. Nobody's visibility changes retroactively: with 0 contributions linked, no account displays a badge today.

## Also here

- **`GET`/`POST /me/supporter/claim-key`** and `/rotate`, shaped like the personal calendar subscription beside it: reading never creates a key, because most accounts never donate.
- **A settings tab** that creates, shows, copies and rotates it, explaining where the key works and where it does not. Rotation is behind a confirmation because it invalidates a key a donor may already have written into a recurring payment. **Ungated** — it appears for every signed-in user on merge.
- **The plan and a research note.** [`docs/findings/selling-software-in-germany.md`](https://github.com/fleetyards/fleetyards/blob/feat/premium-fleet-features/docs/findings/selling-software-in-germany.md) records what selling this eventually requires — VAT, §14 UStG invoices, the consumer-law obligations that carry no revenue threshold, and what a merchant of record does and does not cover. Nothing here builds any of it; the entitlement record is just shaped to accept it later.
- **A locale dedup**, in its own commit: `en`, `zh-CN` and `zh-TW` each declared `addBundledVehicles` and its hint twice, byte-identical. Any JSON load-and-dump drops a copy silently. Noticed in #4912.

## Not in this change

- **Any premium gating.** The entitlement record, enforcement and the four capabilities are Phases 5–7 of the plan.
- **A PayPal or Buy Me a Coffee endpoint.** Hand-entered with the key pasted from the payment note, which is the same resolution path and no new code.
- **Patreon OAuth.** The email match covers Patreon automatically and would do nothing for the other three platforms.

## Testing

| | |
|---|---|
| Ruby | 121 tests, 314 assertions — the key alphabet and its ambiguities, the linker across confirmed/unconfirmed/key-outranks-email/late-signup, the Ko-fi importer over replays, foreign currency, zero amounts and subscription payments, the webhook over bad, missing and unparseable tokens |
| Frontend | 4 new vitest specs; `eslint`, `prettier` and `lint:ts` clean |
| Schema | Regenerates byte-identical; `oasdiff` clean on all three after ignore entries for the admin `kofi` enum widening |

## Two ops steps this cannot do

1. The Patreon access token needs reissuing with **`campaigns.members[email]`** — without it the field is simply absent, which is indistinguishable from "no patron matched". The importer now logs that distinction rather than leaving it to be discovered.
2. A **Ko-fi verification token** in credentials under `kofi.verification_token`. Until it is set the endpoint refuses everything, which is the safe direction.

And the three existing patrons do not need any of it: `Elfwyn` and `RikonGB` match a username exactly and `Christopher Jackson` is a legal name, so they are linked by asking and one admin edit. A name match is deliberately *not* offered as an automatic rule — three people is a conversation, not an algorithm.

🤖

## Summary by CodeRabbit

* **New Features**
  * Added Ko-fi payment support with secure webhook processing and automatic contribution recording.
  * Added supporter claim keys for linking PayPal, Ko-fi, and Buy Me a Coffee payments.
  * Added automatic matching by claim key or confirmed payer email.
  * Added claim-key API access and copy controls in the support modal.
  * Added supporter tiers to public user information.

* **Updates**
  * Patreon imports now use member email addresses for account matching.
  * Anonymous contributions remain included in totals but are excluded from progress lists.
  * Added payer email visibility and expanded translations.


mortik:
All five were valid and all five are fixed in 5df4deb0e. Three of them were the plan promising something the code did not actually do, which is the most useful kind of finding.

**Manual payments remain unlinked** — correct, and worse than described: `payerEmail` also had to be declared on `SupporterContributionInput`, which sets `additionalProperties: false`, so the admin UI could not have sent it even if the controller permitted it. Now permitted, declared on both the input and response components, emitted by the jbuilder, and `Supporters::Linker` runs after a successful create or update. It cannot overwrite an account an admin chose — the linker returns early on a contribution that already has one, and there is a test for exactly that.

**Late signups stay unlinked** — correct. Now an `after_commit` on `User` keyed to `saved_change_to_confirmed_at?`, rather than Devise's `after_confirmation`, so an account confirmed by assignment — an OAuth signup, an admin, a backfill — is covered as well as one that went through `confirm`. Reconfirmation runs it too, which covers somebody moving their account to the address they donate from.

**Former patrons use stale email** — correct on both halves. `end_or_skip` now refreshes `payer_email` from the member before linking, and links after the save rather than before it, matching `upsert`. Two tests: a row with no address picks one up and links, and a row whose patron changed address does not keep the old one.

**Null response violates schema** — correct, and the suggested fix is the house convention (`commodity.rb`, `friendship.rb` and others already use `type: [:string, :null]`). Applied. The generated client now types it `string | null`; the settings page already coalesced with `?? ""`, so no frontend change was needed.

**Concurrent creation invalidates keys** — correct. `ensure_claim_key!` takes a row lock and re-reads inside the transaction.

Worth recording that my first version of the race test was worthless: it passed three times in a row with the lock removed, because each thread called `User.find` *inside* the thread and the second read picked up the first's write, so there was never a race to lose. It now loads both instances before either thread starts, which is what two concurrent requests actually hold. Verified: green with the lock, fails 3/3 without it.

904 tests, all linters clean, `oasdiff` clean on all three schemas.

One thing deliberately left alone: `ensure_calendar_feed_token!` on both `User` and `Fleet` has the identical check-then-write race. Out of scope here, but it is the same bug in the same shape if anyone wants it.

🤖

PR #4917, opened 2026-09-15
feat(patreon): connect a Patreon account to claim its pledges

mortik:
A supporter can connect their Patreon account, and their pledges link to it.

Follow-up to #4915, which left one case unsolved: a patron whose Patreon address is not their Fleetyards address had no self-service path at all. The email match only links those two when they happen to be the same, and the claim key cannot reach us through Patreon — its API exposes no patron-written field, which is why the key exists for Ko-fi, PayPal and Buy Me a Coffee instead.

## The id was already in flight

`Patreon::Client#normalize` resolved the included user record to read its `vanity`, then discarded the `id`. That id is the Patreon **account**, as distinct from `patreon_member_id`, which identifies the *membership* — and it is exactly what an OAuth connection proves ownership of.

It is kept now and stored on the contribution, written by both importer paths. Not unique: one account can hold several memberships over time, and each ended pledge keeps its own row.

## Deliberately not a "my Patreon email" field

The obvious cheaper version is a second address on the account. It is a security hole: an address somebody types is an **assertion**, so anyone could claim a stranger's contribution — and, once premium lands, their entitlement.

What makes the existing email match safe is that neither side is asserted — Patreon verified the address for billing, Fleetyards verified it at confirmation. Signing into Patreon keeps exactly that property. Making a typed address safe would mean a second confirmation flow, which is more work than OAuth for a worse result.

So `Supporters::Linker` gains a third arm, ranked **first**:

| | |
|---|---|
| connected Patreon account | proof of ownership — nobody asserts it |
| claim key | a deliberate act, but can be mistyped into the wrong payment |
| verified email | two verified records, but an address can be shared |

## Linking runs both directions

- **Connect after a sync** — an `after_create_commit` on the connection claims contributions already stored under that account, the way `backfill_discord_member_roles` already does for Discord.
- **Connect first** — the next import links on the way in.

Both have tests, as does the case that must not happen: a connection never takes a contribution somebody already holds.

## No gem, and why

The only published `omniauth-patreon` is **1.0.2, last released in 2018** — predating both OmniAuth 2.x (which this app runs) and Patreon's v2 API. Every other provider here comes from a gem, and this is a deliberate departure from that pattern: an eight-year-old strategy against a v1 API is the worse risk. Patreon's flow is unremarkable OAuth2, so the strategy rides `omniauth-oauth2`, which is already installed and is what the provider gems are built on.

Nothing else needed inventing. Font Awesome already carries `fa-brands fa-patreon`, so the default icon branch covers the button, and `SocialLogins` derives its list from `OauthBtnProvidersEnum` filtered by `oauth-${provider}` — so the enum value plus the flag is the whole frontend change.

## Inert until configured

The devise entry is guarded on both the client id and the secret being present, so this ships dormant. Two ops steps, neither of which a code change can do:

1. Register the redirect URI on the Patreon client — and confirm its **Client API Version is 2**, which is what defines the scope set (there is no scope picker in the portal).
2. `patreon.oauth_client_id` and `patreon.oauth_secret` in the environment credentials. Note these are per-environment files — `config/credentials/production.yml.enc` is read *instead of* the shared one, never merged.

## Testing

| | |
|---|---|
| Ruby | 1661 tests green — 27 Patreon, 13 linker, 8 connection, plus the pre-existing enum-shape assertion updated for `patreon => 6` |
| Frontend | 1527 tests, `eslint`, `prettier` and `lint:ts` clean |
| Schema | regenerates byte-identical; the new flag widens `FeatureFlagName`, which is why the client is regenerated |

🤖

## Summary by CodeRabbit

- **New Features**
  - Added Patreon as a sign-in and account-connection provider.
  - Patreon supporter contributions can now be matched to user accounts using Patreon account IDs.
  - Matching occurs automatically when a Patreon account is connected, with fallback matching options retained.
  - Added a feature flag for Patreon connections and contribution matching.
  - Prevented duplicate Patreon account identities from being connected.

- **Localization**
  - Added Patreon connection, disconnection, and provider labels across supported languages.


mortik:
Both looked at. One taken, one declined with a reason.

**Clear the Patreon identity when the relationship is absent** — taken, in 80c8e7db3. Assigned unconditionally at both sites now.

Worth spelling out why this one is different from the `payer_email` guard sitting next to it, since they look like the same line: the address only helps *find* an account, but `patreon_user_id` decides who may **claim** the contribution by connecting. Holding an identity the current sync does not confirm is worse than holding none, and the two failure modes are not symmetric — a transient omission now costs a sync's worth of connect-claiming and repairs itself on the next run, while a stale id could hand the pledge to the wrong account outright. Test added for the clearing case.

**Keep Patreon unavailable until credentials are configured** — declined, because it is not this PR's behaviour. `oauth-bluesky` already ships exactly this way:

```ruby
if bluesky_private_key.present? && bluesky_jwk.present?
  config.omniauth :atproto, ...
end
```

…with `oauth-bluesky` declared `permanent: true` in the registry, toggleable from the admin UI, and the button rendered off the flag alone. Enabling it without credentials produces the same unregistered-strategy failure the finding describes. So the gap is real but pre-existing and repo-wide, and gating only Patreon would leave the two providers behaving differently for no reason a reader could infer.

The proper fix is to report provider availability — credentials configured *and* strategy registered — rather than inferring it from a flag, and to have `SocialLogins` filter on that for all seven. That is a change to shared OAuth plumbing with its own tests, and it belongs on its own rather than riding a Patreon PR.

Mitigation in the meantime is that the only person who can enable the flag is the same person who configures the credentials, and the failure is immediate and visible rather than silent.

🤖

PR #4929, opened 2026-09-15
fix(logistics): an image with no sized variants is still the picture

mortik:
Follow-up to #4921, which merged a few minutes before this landed. Greptile raised it on that PR and it is worth fixing rather than carrying.

## The bug

`api/v1/shared/_file` emits `smallUrl` / `mediumUrl` / `largeUrl` / `xlargeUrl` only when the attachment is representable, or when it is a vector (where all four are the same file). Anything else is serialized with its original `url` and nothing more.

Both readers of that payload asked for a variant and treated its absence as "no image":

| | asked for | result when only `url` exists |
| --- | --- | --- |
| `useInventoryImage` | `mediumUrl` | panel draws a placeholder |
| both inventory edit modals | `smallUrl` | field shows the fallback preview |

So an upload that succeeded, and that the API was serving, was drawn over with placeholder art in both places.

## The fix

The panel falls through `mediumUrl → url` before reaching for a preset, and a ship's artwork resolves the same way — a hold is still shown as its ship when the ship's picture has no variants either.

The edit field has three cases rather than two, which is what made this easy to get wrong:

```
sized attachment  -> nothing passed; FormFileInput draws it from smallUrl
original only     -> pass the original, because the field cannot find it
nothing attached  -> pass whatever the panel would fall back to
```

`InventoryPanelRecord["image"]` now types `url` alongside the variants, since it is the only one guaranteed to be there.

## Verification

1590 frontend tests — three new, covering an original-only inventory image, an original-only ship image, and a sized variant still winning over the original. `lint:ts`, eslint and prettier clean.

🤖

## Summary by CodeRabbit

- **Bug Fixes**
  - Inventory and vehicle images now display correctly when only an original image is available.
  - Sized image variants are prioritized when available, preventing unnecessary fallback imagery.
  - Inventory creation forms no longer show fallback previews when no image is attached.
  - Image previews now behave consistently across fleet and hangar logistics views.

- **Tests**
  - Added coverage for original-only images, sized-image prioritization, missing images, and inventory creation previews.


mortik:
Both reviews landed on the same gap, and it was the right one: the composable was tested and the modals were not, while the branch this PR fixes lives in the modals. Each of them independently decides between a sized attachment, an original with no variants, and no attachment at all — so every one of those could regress with the composable tests still green.

One spec per modal now, one test per case, plus the create path that has no inventory to stand in for. The sized-attachment test also asserts the file reaches the field, so "no preview" means "the field draws it" rather than "nothing is shown".

`inventoryDefaultImage` is a function declaration now, per `.cursor/rules/frontend.mdc:75`. And the comment the three-case block replaced was still sitting above it saying there were two; that is gone.

🤖

PR #4930, opened 2026-09-15
feat(forms): the shipped art is picked inside the upload field

mortik:
> Stacked on #4926 — review that first; this PR's base is `feat/draft-missions-create-then-edit`.

## Why

A preset was offered in a grid **above** the field it was a value for, and a picked one was then drawn nowhere: the control underneath stayed an empty dropzone. So the form showed everything except the picture the record was about to carry, and the only place the choice became visible was after saving.

The preset is the picture now, and it is shown where the picture goes.

| | before | after |
| --- | --- | --- |
| where the art is offered | a grid beside the field | a modal the field opens |
| where a picked preset is drawn | nowhere | in the field, like an upload |
| what is offered | the record's own type only | everything, opened on its type |
| upload + preset together | the form coerced one away, per form | the field keeps them exclusive, both ways |

## Choosing freely is the point

The chips narrow the grid; they never restrict the choice. A salvage run at a mining site wants the mining picture, and there was no way to say so before.

That cost `FleetContract` half a rule. It used to re-default **any** preset naming a kind the contract no longer was, on the grounds that the form could not have offered that art deliberately — it can now. Only the old kind's own default is replaced, which is the case the rule was actually written for; anything else is a picture somebody picked, and rewriting it would discard a choice silently. Two tests changed to say so, one is new.

## Inventories

They gain `image_preset` on both sides (`inventories`, `fleet_inventories`) and open the same picker. **Undivided on purpose** — an inventory is a place, not a kind of job, so there is nothing to narrow by and every picture is offered at once.

Two holds never take one: a hold that carries an upload shows the upload, and **a ship's hold shows its ship** — which is why a ship's hold is not editable and offers no picture to pick in the first place (`:editable="!record.vehicle"`, already the case).

`app/frontend/images/inventories/` still holds only the two placeholders plus the backdrops the app lends it. Dropping art in there is still the whole of adding a picture — the README says so, and no code changes when you do.

## One catalogue, not four

Three folders of shipped art carried three copies of the same glob, the same format-priority table and the same "`mining_alt1` belongs to `mining`" rule. A fourth caller was about to make it four. `usePresetImages` owns that; the cover composables keep only what is theirs (resolution order, the fleet's own contract covers, the mission art a kind borrows). Net −270 lines across the three.

## Two details worth a look

**`presetSrc` never outranks the attachment.** Every reader of these records prefers the upload, so a preset drawn over one would show a picture that appears nowhere else. It does outrank `previewSrc`, which is a stand-in rather than a choice.

**The picker brings its own overlay**, teleported to the body on AppConfirm's layer rather than going through `AppModal`. The app has one modal, and this control sits inside one in both logistics forms — going through it would have replaced the form being filled in. Escape is taken at the capture phase and stopped there so the surface underneath does not close as well.

**Clearing works from the field.** The clear button only appeared for a file, so a field showing a preset had no way back to no picture except through the picker. It appears for either now, and clears both — the button means "no picture".

**The picker button is always visible**, unlike the clear button beside it. An offer that only appears under the pointer is one nobody on a touch screen is ever made. Opposite corner from clear, too: the two are not a pair.

## One thing that is not mine

`ruby-tests` failed on the parent branch and on this one with the same seven failures: `ValidationError.new("missions.publish", …)` has no translation in any locale, so a refused publish rendered `translation missing: en.validation_error.missions.publish` into the payload. It belongs to #4926, which added the endpoint — but that branch is checked out in another worktree, and the whole stack stays red until somebody writes the seven lines. They are here. If the parent adds them too, the two copies are identical.

## Verification

1619 frontend tests (33 new, across the catalogue, the modal and the field), 256 inventory/contract Ruby tests, `lint:ts`, eslint, prettier, standardrb, knip clean. Schema regenerated — 28 lines, all `imagePreset`; `swagger/admin/v1` untouched. `oasdiff` against the parent branch reports no breaking changes. Migration produces a 4-line `db/schema.rb` diff. New states added to `/visual-tests/forms` for both halves of the field.

🤖

## Summary by CodeRabbit

- **New Features**
  - Added image preset selection for inventory, mission, and contract images.
  - Added searchable, filterable preset picker with selection, removal, keyboard, and accessibility support.
  - Inventory image presets can now be saved, cleared, and returned through the API.
  - Added localized preset controls and labels across supported languages.

- **Bug Fixes**
  - Preserved custom artwork when changing contract types.
  - Improved image fallback behavior and retained uploaded images appropriately.
  - Refined edit-page tab recognition and titles.


mortik:
@greptile-apps review

mortik:
Fixed, and the finding was right: declaring `aria-modal` is a promise that everything behind the dialog is inert, and the keyboard was not keeping it — the form underneath stayed tabbable, focusable and invisible.

Three parts, in `c465049fd`:

- **In.** Focus lands on the dialog element itself on open, not on the first tile. It carries the `aria-label`, so a screen reader says what this is before reading out a grid of pictures; `tabindex="-1"` lets it hold focus without joining the tab order. `role`/`aria-modal` moved from the scrim onto the panel while I was there — the dialog is the panel, not the backdrop behind it.
- **Contained.** Tab wraps at both ends, and is pulled back to the first control if focus is ever found outside the dialog. Handled in the same capture-phase listener that already owns Escape.
- **Back.** The field restores focus to the button that opened the picker, because the field owns that button. Without it, closing drops a keyboard onto `<body>` — the top of the page.

Four tests on the trap and one on the restore. The restore test needed `mountWithDefaults` to gain an `attachTo` option: `focus()` does nothing for an element that is not in the document, so a focus assertion without it sees `<body>` and passes or fails for the wrong reason.

Also rebased onto `main` now that #4926 has merged, and absorbed #4931. That one added a `detachedForPreset` flag to tell the form's own detach from the author's clear — there is nothing to tell apart here. Removing a preset restores nothing, so there is only ever one `null` and it always means what the field is showing: no picture. Both of #4931's tests are kept and pass against the simpler model.

🤖

PR #4932, opened 2026-09-15
fix(logistics): a ship's hold is its ship, and four other review leftovers

mortik:
The review findings that arrived on #4930 after its last check and merged unaddressed, plus the oversized placeholder.

## Two that change what you see

**A ship's hold drew a crate over its ship.** The rule is that a hold riding in a ship is shown as that ship — it is why no picker is offered for one — but the resolution order put a chosen preset above the ship image. Not reachable through the app today (`vehicle_inventory` has only `show`/`destroy`, and the hangar endpoint addresses hand-made inventories only), so nobody would have hit it; the column takes a preset regardless, and the rule belongs in the resolution rather than in what the UI happens to allow.

**A contract's default preset outranked the fleet's own cover.** `FleetContract` writes the kind into `cover_image_preset` whenever the form left it empty, so nearly every contract carries one — and resolution read that as a choice, returning shipped art ahead of the cover a fleet configured for that kind in its settings. Only a preset naming something *other* than the kind counts as a choice now, which is exactly the set the picker produces.

That one predates the picker — the ordering is unchanged from before it — but the picker is what made a cross-kind preset expressible, so the distinction is worth drawing now.

## Two that do not

`stopPropagation` → `stopImmediatePropagation` on the picker's Escape handler. The former stops the later phases but not another listener on the same target, and `window` is exactly where the app modal and the confirm dialog listen.

The test it had proved nothing — it dispatched at `window` and registered its probe after mounting, so it passed with either call. It presses Escape from inside the dialog now, with the probe registered before the picker mounts and in the bubble phase, which is how the surfaces underneath actually listen. **That version fails without the call**; the old one did not.

## The schedule tab and its title disagreed

| | tab | title before | title now |
| --- | --- | --- | --- |
| es | Programación | Editar **fecha** | Editar programación |
| fr | Planification | Modifier **la date** | Modifier la planification |
| it | Programmazione | Modifica **Data** | Modifica Programmazione |

That tab holds the date, the time and the recurrence. Three locales titled the document after one of the three. Matched to the tab rather than merely widened, so the two are the same word wherever both exist.

## And 1.1 MB of placeholder

`placeholder-3.webp` shipped as a near-lossless encode — 1.34 MB for 2000×1125, where its own jpg is 360 KB at the same size and `placeholder-1` is 493 KB for twice the pixels. The catalogue prefers webp, so the heavy one is what a panel background actually loaded.

Re-encoded at q90 **from the webp rather than the jpg**, so no jpg artefacts are baked in twice: 210 KB, and 91 KB per megapixel against `placeholder-1`'s 103 — the same density as the art already here.

## Verification

1651 frontend tests (5 new), `lint:ts`, eslint, prettier clean. No schema, migration or Ruby changes.

🤖

## Summary by CodeRabbit

- **Bug Fixes**
  - Contract covers now distinguish between an unset preset and an explicitly selected preset, preserving intentional selections and applying fleet-configured covers when no preset is chosen.
  - Inventory images now prioritize the inventory’s own image, followed by vehicle imagery, selected presets, and default artwork.
  - Prevented the image picker’s Escape key from unintentionally closing underlying dialogs.

- **Localization**
  - Updated event scheduling titles in Spanish, French, and Italian for clearer wording.


mortik:
Right, and it exposes that the previous commit was the wrong shape of fix.

Two reviews now disagree about the same value, and **both are correct**:

- CodeRabbit on #4930: a preset equal to the kind must **not** outrank the fleet's configured cover — because the model wrote exactly that whenever the form left the field empty.
- This one: it **must** outrank it when somebody picked that tile deliberately — which the picker offers, since it shows every kind's art including your own.

`"crafting"` on a crafting contract meant either thing and no reader could tell which. Guessing in the reader is what I did last commit, and you're right that it discards a real choice.

So the ambiguity is gone from the source instead: **nothing writes a default any more.** An unset preset is the absence of a choice, which is what lets the fleet's cover answer; a stored one is a picture somebody picked, whatever it names. The re-defaulting on a change of kind goes too — there is no stale default left to clean up after.

Safe to do rather than paper over:

- **Nothing else reads the column.** No OG image, no push, no server-side render — only the frontend's own resolution, which already fell back to the kind's art when the preset was absent. So `nil` resolves to the same picture it did before, just after the fleet's cover rather than before it.
- **No data to migrate.** `FleetContract.count` is 0 in the production dump — the feature has never been used, so there is no ambiguous row and no choice to lose.
- The column was already `[:string, :null]` in the schema, so the generated client and swagger are untouched.

Five tests changed to say what the column now means, including the one that matters: a preset naming the contract's own kind survives, which is the case your comment is about.

1652 frontend tests, 29 contract model tests, 27 contract integration tests, lint and standardrb clean.

🤖

PR #4991, opened 2026-09-17
chore(sc_data): bump live to 4.10.1-live.12660092, and check a parsed tree before it ships

mortik:
Live moves to `4.10.1-live.12660092`.

## The build

A patch: one file leaves the export, 124 arrive. The AEGS Sabre Raven EX with its own armour, thrusters, seat, dashboard, tail and winglets; the ARGO ATLS Ikti-NTLS and the repeater it carries; four sets of armour (KAP heavy, QRT specialist, VGL undersuit and backpack, a DOOM undersuit helmet); six paints with their swatch records; five store variants of weapons already in the catalogue; and the crafting quantization records for 22 gatherables. The parsed tree gains 48 files for all that and loses none.

The export itself is complete, and it was worth checking: the bucket lists 77,323 objects / 3,888,281,827 bytes and the synced tree matches it byte for byte, the record directories are the same 3,647 as 4.10.0, and all 62,434 record XMLs parse.

## Thirty-six paints that never had a swatch

Looking at the tree turned up a bug older than this build. Those paints name their artwork with Windows separators — `ui\sharedassets\paintcolorlogos\paint_aurora_black_red_white_icon.tif` — while every other record in that folder uses slashes, and `raw_assets` is keyed on paths the file system produced. The lookup missed, so nothing was copied, and the path the record kept was one `Dir.glob` could not match either. The art was in the export the whole time; normalising the separator on both sides recovers 35 files (two paints share one swatch).

## A check for the parsed tree

Every way a parse goes wrong is quiet. A half-finished upload parses into a tree that is merely small; a category the export renamed into a folder that is merely empty; the separator above into a path that merely never resolves. None of them raise — the tree loads, and what it no longer carries is retired days later without anyone deciding that.

`ScData::ParsedCheck` asks what a person asks after a parse: does the tree claim the build the config points at, is every catalogue present and above a floor a build could not honestly fall below, does every record parse and carry the field that identifies it, does every icon a loader will look for resolve, is anything zero bytes. It reads the tree and nothing else — no network, no database — so the same code answers from `bin/scdata check <env>`, at the end of a `parse`, and as a test over the tree CI pulls.

Only `push` acts on it, since that is where a bad tree stops being local and becomes the one the loaders pull. `--force` is there for the day a real build lands under a floor.

Drift is deliberately not judged: a build that drops eleven ships has dropped eleven ships, and no threshold tells that from a mistake. `ScData::BuildCompare` answers that against loaded rows, where both builds are present and the list can be read.

```
  commodities       232 files, 63/63 icons resolve
  equipment        4862 files
  items            7795 files, 929/929 icons resolve
  manufacturers     118 files, 109/110 icons resolve
  models           1111 files
  Nothing to report.
```

The one manufacturer logo that does not resolve is art the export does not ship — that is counted rather than failed, up to the point where the count stops describing records and starts describing a missing folder.

## Before merging

The parsed tree has to be in the bucket before this pointer lands, or a loader pulls the previous build for a version that names this one.

🤖

## Summary by CodeRabbit

- **New Features**
  - Added a `check` command to validate parsed catalogue data and report detected problems.
  - Parsing now reports validation results, while publishing blocks invalid data unless `--force` is used.
  - Asset paths are normalized consistently for icon loading and copying.

- **Configuration**
  - Updated the live source data version to `4.10.1-live.12660092`.

- **Bug Fixes**
  - Improved detection of malformed JSON values, missing catalogues, invalid records, metadata issues, empty files, and missing or incorrectly referenced icons.
  - Empty-file validation now reports all detected problems.


mortik:
Thanks both — the first two are right and are fixed in ff38cf934.

**Non-object JSON** (`parsed_check.rb`): a file holding valid JSON that is not a record — `null`, a bare list — got past `JSON.parse` and then took every key it was asked for as a method call on nil or an index into a list, so a check whose whole purpose is to report a broken tree raised from the middle of one. It now names the file and what it holds, for catalogue records and `version.json` alike, and carries on through the rest. Two tests cover it.

**Truncated empty-file list**: agreed — where to stop printing is the caller's decision, and `report_check` already makes it. The checker returns every empty file now, so the count is the truth; a test writes twelve and asserts twelve.

**`--force` scope**: adopted in part. The real problem was that `--force` skipped the check rather than the verdict, so the push most worth having a record of printed nothing — it now always runs and always prints, and says when it was overruled. The override itself stays a blanket one, though. A tree can be unfit for reasons nobody foresaw, and an operator who has just read the whole list is better placed to judge that than a severity ranking written months earlier; a partial override would also send someone straight to editing the check out when the one case they need is on the wrong side of the split.

🤖

mortik:
On publishing the pointer only after the upload — not taking this one, for three reasons.

A parse does not publish anything. It rewrites a working-tree file; the pointer becomes real when a commit carrying it is merged and deployed, which is a separate act by a person, and on this branch the tree went into the bucket before the PR was opened.

The failure it describes is already guarded by the thing it cites: `FetchesParsedTree` calls `ParsedStore#verify_remote_version!`, so a pointer ahead of the remote tree makes the loader raise `VersionMismatch` and refuse. That is loud and correct — nothing loads the wrong build or mis-stamps rows.

And deferring the config write would couple two commands that are deliberately separate. `parse` needs no credentials and no network; `push` needs both. Someone parses, reads the check, and pushes later — or never, for a ptu build they only wanted to look at. Writing the config only on a successful push would leave the parsed tree on disk describing a build the config does not name, which is the same mismatch pointing the other way, with nothing to verify it.

🤖

PR #5001, opened 2026-09-17
fix(sc_data): parsers keep a name declared with a plural marker or in another case

### LINKED ISSUE #4997: sc_data parsers drop a name declared with a plural marker or in another case
## Description

`ScData::Parser::BaseParser#translate` matches a localisation key **exactly**. The game declares a large number of keys in another case, or with a `,P` plural marker that the record referring to them does not carry — so the lookup returns `nil` and the record loads nameless, or is dropped.

Measured against `4.10.1-live.12660092`:

| | effect |
|---|---|
| **84 equipment records** | `EquipmentParser#parse_equipment` returns early on a blank name, so these are **absent from the catalogue entirely** |
| **234 ship item records** | `ItemsParser#parse_item` keeps the row with `name: nil` |
| all 73 `@crafting_ui_slotname_*` | every crafting slot label |
| 24 of the 29 `@StatName_GPP_*` | every crafted-property name bar five |

```ruby
key = "@item_Name_qrt_combat_medium_arms_01_02_01"
translate(key) # => nil        -- global.ini declares "...,P"
localize(key)  # => "Testudo Arms Purgatory Camo"
```

Examples of what is missing today:

- equipment: `utfl_crossbow_ballistic_01_tint01` → Novian "Nighthunter" Crossbow, `behr_rifle_ballistic_01_sf01` → P4-AR "Star Kitten" Rifle, `qrt_combat_medium_arms_01_02_01` → Testudo Arms Purgatory Camo, `mrai_flightsuit_01_07_02` → Neutrino Racing Flight Suit WhiteHot
- ship items: `armr_aegs_redeemer` → Aegis Redeemer Ship Armor, `armr_rsi_polaris` → RSI Polaris Ship Armor, `controller_flight_aegs_eclipse` → Flight Blade

## How it surfaced

While building the blueprints catalogue (#4988). 23 of the 1607 crafting recipes make an item that resolves to no `Equipment` row — every one of them a colourway or a tint whose name is declared with a plural marker. The recipes load and link to nothing.

## What already exists

`ScData::Parser::CommoditiesParser` has always worked around this with its own `,P`-stripped, downcased index, which is why commodities do not have the problem. #4988 lifted that index onto `BaseParser` as `normalized_translations` / `localize`, and `CommoditiesParser` now builds on it — verified byte-identical over all 232 commodities. So the fix is available; it is simply not used by the other three parsers yet.

## Acceptance Criteria

- [ ] `EquipmentParser`, `ItemsParser`, `ModelsParser` and `ManufacturersParser` resolve a name through `localize` rather than `translate`
- [ ] The 84 equipment records load; the colourways and tints among them keep `hidden: true`, which `VARIANT_SUFFIX` already decides separately from the name
- [ ] The 234 ship items carry their name
- [ ] Re-parse and `bin/scdata push` in the same PR, with a snapshot diff proving the change is only the names
- [ ] A parser test per parser covering a key declared with `,P` and one declared in another case

## Notes

- Deliberately **not** done inside #4988: it rewrites the equipment and items trees, and that wants its own review and its own re-parse rather than riding along with a new catalogue.
- `translate` itself should stay. It is right for a key spelled the same on both sides, and `describe` parses the description prose it returns.
- Check whether any name that appears twice after the fix collides on a slug — `EquipmentParser#mark_skins` groups by name, so 84 new names can change which record in a family stays visible.

🤖

### D1 — A name the export marks as a placeholder is not a name

25 of the equipment records the index unlocks resolve to `PH - hdh_boots_01_01_13`,
`PLACEHOLDER - SPV Jacket` or bare `PH`, and 34 item names do the same. **None of
them exist in the parsed tree today** — they only start resolving once the index
is consulted, so shipping them would be a regression introduced by the fix rather
than a name recovered by it.

`localize_name` drops a name matching `/\A(ph|placeholder)(\s|\z)/i`. Across the
7343 distinct names 4.10.1 resolves, that matches 26 and no real name — it is
anchored to a leading marker word, so "PH-13 Phase Rifle" keeps its name.

This is why the landed counts are **59 equipment and 200 item names**, not the
84 and 234 measured above: those figures are gross of the placeholders.

### D2 — GEND and GENF are skipped, not loaded

Both resolve to the same string, "Consumable". `ManufacturersLoader` matches a
manufacturer on the slug its name produces, so loading both mints a single
manufacturer called "Consumable" holding both codes.

They are the generic maker the export files a carryable under when it has no real
one; their logos (`GenericDrink_256.tif`, `GenericFood_256.tif`) say what they
stand in for. The 107 drink and 171 food records pointing at them are props and
carryables that no catalogue here scans, so nothing loses its maker by dropping
them. Both are now `skip: true` in `config/sc_data/manufacturer_overrides.yml`.

### D3 — Names only; descriptions stay on `translate`

Read through the index a description arrives with its `\n` markers already turned
into real newlines — a different string from the one the export writes and from
the one all 3019 affected descriptions record today, and `EquipmentParser#describe`
parses the prose `translate` returns. Career and role resolve exactly for every
key 4.10.1 declares, so they gain nothing and were left alone.

### D4 — The re-parse and push wait for #4988

PR #5001 is stacked on #4998, which was still moving. Parsing from this branch
would have overwritten the shared tree with output from a blueprints parser older
than the one #4998 now carries. The rebase-onto-final-tip, re-parse and
`bin/scdata push` happen once, before merge.

🤖

mortik:
Resolves #4997

`ScData::Parser::BaseParser#translate` matches a localisation key exactly. The export declares a large number of keys in another case, or with a `,P` plural marker the record referring to them does not carry, so the lookup returned `nil` and the record loaded nameless — or, for equipment, was dropped.

`normalized_translations` / `localize` came from #4998, which has since squash-merged as `797e259`; this is now rebased onto `main` and stands alone.

## What changed

`localize_name` on `BaseParser` reads the shared normalized index and falls back to `translate`. The fallback is what makes this a safe swap: whatever `translate` resolved, this still resolves, and the index only adds the keys it could not match. `EquipmentParser`, `ItemsParser`, `ModelsParser` and `ManufacturersParser` now read a display name through it.

Names only. Career and role resolve exactly for every key 4.10.1 declares, so they gain nothing, and descriptions stay on `translate` — read through the index a description arrives with its `\n` markers already unescaped, a different string from the one all 3019 of them record today, and `describe` parses the prose it returns.

## Two things the issue did not anticipate

**Placeholder names.** 25 of the records the index unlocks are named `PH - hdh_boots_01_01_13`, `PLACEHOLDER - SPV Jacket`, or just `PH`. None exist in the tree today. `localize_name` drops a name matching `/\A(ph|placeholder)(\s|\z)/i` — across the 7343 distinct names 4.10.1 resolves, that catches 26 and no real name.

**GEND and GENF both resolve to "Consumable".** The loader matches a manufacturer on the slug its name produces, so loading both mints one manufacturer holding both codes. They are the generic maker the export files a carryable under; the 107 drink and 171 food records pointing at them are props no catalogue here scans, so both are now `skip: true` in the overrides — and in `DedupeManufacturersTask::DROPPED_CODES`, which a test asserts agrees with them, so a fresh import and a cleaned table cannot disagree about what a code is. That task still refuses to drop a row anything points at.

## Snapshot diff

Parsed twice over `4.10.1-live.12660092` in an isolated sandbox, before and after. No record lost, no field cleared, nothing but names:

| folder | records | change |
|---|---|---|
| equipment | 4862 → 4921 | **59 added**, 0 removed, 0 fields changed on an existing record |
| items | 7795 → 7795 | **200 names** filled, 9 short names (4 filled, 5 unescaped†) |
| models | 1111 → 1111 | **9 names** filled, 1 rewritten (`Argo CSV-SM\n` → `Argo CSV-SM`) |
| manufacturers | 118 → 116 | 1 name filled (CCC), GEND and GENF skipped |

† Five records point their `ShortName` at a `_Desc` key, so they have carried a
whole paragraph as a short name for as long as the tree has existed — the two
Storm turrets and three mining modules. Reading a name through the index
unescapes their `\n`, which is the entire change to those five. Pre-existing, and
left alone here.

Examples: `armr_aegs_redeemer` → Aegis Redeemer Ship Armor, `armr_rsi_polaris` → RSI Polaris Ship Armor, `behr_rifle_ballistic_01_sf01` → P4-AR "Star Kitten" Rifle, `aegs_sabre_firebird_collector_milt` → Aegis Sabre Firebird Wikelo War Special.

The numbers are lower than the issue's 84 and 234 because the placeholder guard holds 25 equipment records and 34 item names back.

## The checks the issue asked for

- **`mark_skins` fallout** — 0 records flipped `hidden`. The 59 new records arrive 8 hidden, 51 visible; `VARIANT_SUFFIX` decides that from the key, so a record gaining a name does not gain visibility.
- **Slug collisions** — the set of colliding visible-equipment slugs is identical before and after (20 sets, 40 records). No new collision.

## Tests

A parser test per parser covering a key declared with `,P` and one declared in another case, plus the placeholder guard and the description invariant.

`test/parsers test/loaders test/lib/sc_data test/tasks test/services/manufacturers` is 514 green. That last pair matters: `DedupeManufacturersTaskTest` is what catches a `skip:` entry that was not mirrored into the task, and it sits far enough from the parsers that the obvious local selection misses it.

## The tree is pushed

Re-parsed and pushed from this branch once #4998 had landed, so CI and production read the fixed names:

```
Checking the parsed tree (live)...OK
  blueprints       1607 files
  commodities       232 files, 63/63 icons resolve
  equipment        4921 files
  items            7795 files, 929/929 icons resolve
  manufacturers     116 files, 107/108 icons resolve
  models           1111 files
  Nothing to report.
Uploaded: 432, removed: 4, unchanged: 16571
```

The 4 removals are `gend.json`, `genf.json` and their two logos. Counts match the sandbox measurement above exactly — 4921 equipment, 116 manufacturers. The parse rewrote no tracked file: the build is unchanged, so `config/app/sc_data.yml` and `commodity_keys.json` came out identical.

The one manufacturer icon that does not resolve is KAP (`kap_256_diff.tif`), which the export has never shipped a drawable for. Pre-existing and untouched here.

🤖


mortik:
The Greptile review above was run against `15882ba`, this branch's **pre-rebase** tip, and diffed it against `main` rather than against this PR's base. That pulled the whole of #4998 into its view, and all three findings landed there — none of the three files it flagged is in this PR's diff:

| finding | file | in this PR? | status at the current base tip (`8214a80`) |
|---|---|---|---|
| Output lookup ignores source | `app/models/blueprint.rb:90` | no | **already fixed** — `.making` now takes a source and reads `BlueprintBuild.current(source)` instead of the shared columns (`8214a80`) |
| Recipe update is not atomic | `app/lib/sc_data/loader/blueprints_loader.rb:39` | no | **already fixed** — `one` wraps `apply`, `apply_build` and `persist_costs` in a single `Blueprint.transaction` |
| Parser skips hide recipe loss | `app/lib/sc_data/parser/blueprints_parser.rb:97` | no | still silent per record, but `ParsedCheck` carries a `blueprints` floor of 1280 and `bin/scdata push` refuses a tree below it, so a mass skip cannot reach the bucket unnoticed |

This PR changes 11 files, all of them under `app/lib/sc_data/parser/`, `config/sc_data/manufacturer_overrides.yml` and `test/parsers/sc_data/`. Greptile raised nothing against any of them.

Since it has now been rebased onto `8214a80`, a fresh run would diff the 11 files it actually changes.

🤖

PR #5025, opened 2026-09-18
feat(lists): a sort line on the card lists that had no way to sort

mortik:
Column headings are the only sort control the site has. Every list drawn as cards therefore drops
the lot — the API has offered these sorts all along, with nothing in the UI to ask from.

| List | Renders | Sorts the API offers | Reachable before |
|---|---|---|---|
| `/ships/` grid view | `Grid` + `ModelPanel` | 16 | only by switching to the table |
| Hangar + wishlist | `Grid` *or* `VehiclesTable` | 18 | only by switching to the table |
| **Public hangar + wishlist** | `Grid` only | 18 | **no — no table view exists** |
| `/images/` | `Grid` only | 1 | **no** |

## Where it goes

Grid-only where a table view exists, since those headings already carry the same sorts. The whole
control where cards are the only view — which is every public hangar.

Both run through `useTableSorting`, so the chips and any headings read and write the one `q[s]`,
cycle none/asc/desc in the same order, and take their arrows from the same pair. Neither can
disagree with the other about what is active.

## Why not every sort

A line of sixteen or eighteen chips is a wall. These are the ones a card already shows, and the
labels come from each table's own column namespace so both views name a sort the same way — all
of them already translated in all seven locales, checked rather than assumed.

`flagship` is left out of the hangar despite leading its default order: it is a yes/no, so sorting
by it is a filter wearing the wrong clothes.

## Checks

Every chip verified against its model twice over — whitelisted in `ALLOWED_SORTING_PARAMS`, *and*
actually emitting an `ORDER BY`. Ransack drops a sort missing from `ransackable_attributes`
without a word, which is exactly how a chip could look right and do nothing; that bit the
components table's `sizeOrder` earlier in this stack.

`lint:ts`, eslint and the table suites clean.

## Stacked

Targets #5021, which is where `SortBar` lands. Merge that first.

🤖

## Summary by CodeRabbit

- **New Features**
  - Added a searchable Components Catalogue with filters, pagination, sorting, category icons, metrics, badges, and responsive layouts.
  - Added catalogue navigation, localized labels, breadcrumbs, and redirects from legacy component URLs.
  - Added sorting controls across fleets, ships, hangars, wishlists, notifications, payouts, and supporter contributions.
  - Added component links from hardpoints and travel-time views when catalogue pages are available.

- **Bug Fixes**
  - Improved build selection during data updates and preserved event ordering when no sort is specified.
  - Prevented uncatalogued components from appearing in browsable lists.


mortik:
All five valid, and this is the important finding of the branch — thank you. Fixed in ea1d1cc09.

The cause is one thing: `useFilters` strips `s` out of `filters`, so a page building its request
from `filters` drops the sort, and a page sending only pagination never had it. `getQuery()` keeps
it, which is exactly why ships, the own hangar and the fleet ships list worked and the rest did
not. I added the controls without checking the request carried the value.

I audited every page I had touched rather than fixing the five named:

**Wired** — the endpoint declares the sort:

| list | was | now |
|---|---|---|
| public hangar, own wishlist | `q: filters.value` | `q: getQuery()` |
| public wishlist | no params at all | `q` with the sort |
| contracts | `q: { stateIn }` | sort merged in |
| admin contributions | fixed `startedAt desc` | the heading's sort, falling back to that default |

**Removed** — the endpoint has no way to take one, so the control could only ever lie:

| list | why |
|---|---|
| `/images`, ship gallery | `ImageQuery` declares no sort, and the controller hardcodes `order("images.created_at desc")` |
| fleet events | there is no `FleetEventQuery` — `q` is in the controller but not in that path's schema |
| admin inventories | the params carry no `q` |

Those need the endpoint opening up first, which is a schema change per path rather than a line at
the call site. The model-side whitelist for events stays: it is correct on its own and the schema
work builds on it. Happy to do either as a follow-up.

Also fixed `Images.spec.ts`, which counted every `<a>` inside the list — the test id sits on the
`FilteredList` root, so the sort chip counted as an image. It targets the grid now.

🤖

mortik:
@greptileai review

Both branches moved since the last pass — the sort controls that could not reach their endpoint are gone, the rest carry `q[s]`, and the schemas are regenerated. 🤖

PR #5031, opened 2026-09-18
feat(blueprints): the catalogue pages, and the stats a recipe's materials buy

### LINKED ISSUE #4988: Blueprints — a catalogue of what can be crafted, from what, and where the recipe comes from
## Status — data and API shipped, pages not started, 2026-09-18

Three of four phases are merged. What is missing is the visible half: blueprints has a public API
and **no page anywhere**. Plan: `docs/exec-plans/4988-blueprints-crafting-catalogue.md`.

| phase | PR | state |
|---|---|---|
| 1 — parser and loader | #4998 | **merged** |
| 2 — contract sources (pool → generator → org) | #5010 | **merged** |
| 3 — the public API | #5013 | **merged** |
| 4 — the pages | — | **not started** |

**#5017** dropped the `blueprints` feature flag, so the API is live and unguarded.

### D8 is overtaken — the shell was built by #4989, and blueprints is its second tenant

D8 put the shared catalogue shell on this branch, with blueprints as first tenant, and #4989
parked its own page phase on it. That is no longer what happened: this issue's Phase 4 never
started, #4989 waited as long as it could and then built the section itself in #5021.

So Phase 4 no longer builds a shell — it **joins** one:

- the section is `/catalogue/`, which lands on the first tenant that has pages
- `/catalogue/components/` is live, with a list and a page per component
- `/catalogue/blueprints/` is what this issue still owes

The pattern to reuse is already on the branch: `FilteredList` + a `FilterForm` + a row list +
`SortBar` + `Paginator`, with a `useBlueprintFilters` / `useBlueprintSortFields` pair beside the
component ones. Plus a `catalogue/blueprints/:slug` Rails route and a `Frontend::BaseController`
action for meta tags, the way components has one — without it a shared link renders the generic
card.

**A tenant earns its tab when its pages exist, not when its API does.** Blueprints is the live
example of the distinction: the API has been public since #5013 and the nav lists nothing,
because a tab pointing at a route with no page lands on the generic not-found.

### The reverse link now spans four catalogues

The AC below — "a component, equipment or commodity page says which blueprints make it, and which
blueprints consume it" — is unshipped, and the ground under it has moved:

| target | rows | surface |
|---|---|---|
| `Equipment` | 1,103 recipes make one | no pages — **#5027** |
| `Component` | 476 recipes make one | pages live since #5015 |
| `Commodity` | consumed as a cost, via `BlueprintCostOption#commodity` | no pages — **#5026** |

Components is the only one of the three with a page to put the link on today. The other two are
now tracked as their own issues, and each carries the reverse link in its own acceptance
criteria, so this issue owes the blueprint-side half and the component page.

### One trap for whoever picks up Phase 4

**The local dump carries 0 blueprints** — the loader has never run against it, so every page
built against local data renders an empty list until it is loaded. The counts in this body come
from the parsed tree, not from a database.

---

## Description

A searchable, filterable **Blueprints** catalogue: what can be crafted, what it costs in resources, and where the blueprint itself comes from. List with filters, detail page per blueprint, cross-linked to the existing component, equipment and commodity catalogues.

Crafting is in the game files and has been since 4.9.0, so this is a loader-and-surface job, not a research job. Everything below was measured against the raw dump of **4.10.1-live.12660092**, which is what `data/sc_data/raw/` holds today.

## What the game files hold

Everything lives under `Data/Libs/Foundry/Records/crafting/`:

| | count | what it is |
|---|---|---|
| `blueprints/crafting/**` | **1607** | one `CraftingBlueprintRecord` per craftable thing |
| `blueprints/dismantle/globalgenericdismantle.xml` | 1 | the single global dismantle rule (efficiency 0.5, 15s) — not a recipe |
| `blueprintrewards/**` | 154 | `BlueprintPoolRecord`, a weighted list of blueprints |
| `blueprintcategories/` | 21 refs | the category list — see D5 |
| `craftedproperties/` | 24 | the stats crafting quality moves |
| `qualitydistribution/`, `qualityquantization/` | 22 / 63 | how a resource's quality rolls |

A blueprint record is one tier (1607 of 1607 have exactly one `CraftingBlueprintTier`) holding one `CraftingRecipe`:

- `CraftingProcess_Creation entityClass` — **what it makes**, as a record GUID
- `craftTime` — a partitioned time value, 1m35s to several minutes
- `mandatoryCost` → a nested `CraftingCost_Select` tree: an outer select of N named **slots** (`ASPECTS`, count=3), each an inner select of 1 that names the slot (`@crafting_ui_slotname_armouredcarapace` → "Armored Carapace") and lists its `CraftingCost_Resource` options — `resource` GUID, `minQuality`, and a quantity in `SStandardCargoUnit`
- per slot, a `CraftingCostContext_ResultGameplayPropertyModifiers` block: which stat this slot moves and by how much, as a linear ramp over quality 0→1000 (e.g. `modifierAtStart="0.8" modifierAtEnd="1.2"`)

No `optionalCost` exists in any of the 1607.

### What it makes

1604 distinct output refs; 1603 resolve to an entity in the same export:

| output lives under | count | our table |
|---|---|---|
| `entities/scitem/characters/**` (armour, starwear, undersuits) | 916 | `Equipment` |
| `entities/scitem/ships/**` (cooler, powerplant, shield, radar, QD, weapons, mining, salvage, tractor, refuelling) | 473 | `Component` |
| `entities/scitem/weapons/**` (fps weapons 174, magazines 36) | 210 | `Equipment` |
| `entities/scitem/carryables/**` (4 mission items) | 4 | `Commodity` |

By blueprint folder: `fpsgear` 1126 (armour 916, weapons 174, ammo 36), `vehiclegear` 477, `missionitems` 4.

### What it costs

- **26 distinct resources**, all of them refined materials — Iron, Copper, Titanium, Quantanium, Laranite, Bexalite, Pressurized Ice, … Every one already exists as a parsed commodity (`items_commodities_*`), so the resource side is a pure join.
- **11 item ingredients** (`CraftingCost_Item`), all hand-carried harvestables — Aphorite, Hadanite, Janalite, Saldynium ore, a Yormandi eye. All are `entities/scitem/carryables`, which `CommoditiesParser` already reads.

### Where a blueprint comes from

A blueprint is referenced by **nothing outside `crafting/`** except the reward pools. The chain is:

```
blueprint ──> BlueprintPoolRecord (weighted) ──> ContractGenerator ──> mission-giver org
```

Pools split into `blueprintmissionpools` (473 blueprints), `48blueprints` (290), `xenothreat2rewards` (25), `ors` (17), `collectorwikelo` (9). The contract generators that reference them are filed by guild and org — Eckhart Security, Foxwell Enforcement, Headhunters, Shubin Interstellar, Intersec Defense Solutions, Ling Family Hauling — and each carries localised mission titles plus the org's `@*_RepUI_Name`.

**732 of 1608 blueprints sit in at least one pool.** The other 876 have no stated source in the export.

Separately, `entities/basebuilding/interactables/basebuilding_interactables_itemfabricator_{standard,allianceaid}.xml` whitelist blueprint categories per crafting queue, so the category also decides *which machine* can make it.

## What exists today in Fleetyards

| | where |
|---|---|
| `Commodity` (232 rows, `sc_key`/`sc_ref`, builds, `current_version`) | `app/models/commodity.rb`, `ScData::Parser::CommoditiesParser` |
| `Component` (ship items, `sc_key`/`sc_ref`, `hidden` unless its category is listed) | `app/models/component.rb`, `ItemsParser` + `ItemsLoader` |
| `Equipment` (armour, clothing, fps weapons, consumables) | `app/models/equipment.rb`, `EquipmentParser` + `EquipmentLoader` |
| public API | `Api::V1::{Commodities,Components,Equipment}Controller` + `filters/` |
| admin surfaces | `app/frontend/admin/pages/{commodities,components,equipment}` |
| parser/loader registration | `ScData::Parser::BaseParser.all` / `Loader::BaseLoader.all` |

There is **no** crafting data of any kind in the repo — `grep -ri blueprint` finds two unrelated hits.

## Decisions this needs before code

**D1 — A blueprint has no name of its own.** 1606 of 1607 carry `blueprintName="@LOC_PLACEHOLDER"`; the one exception is a mission item. The name, icon and manufacturer all have to come from the crafted output, which makes the output link mandatory rather than decorative, and makes the slug derive from it too.

**D2 — The output is polymorphic.** One blueprint points at a `Component`, an `Equipment` or a `Commodity`. Either a polymorphic `craftable` association (the shape `FleetInventoryItem#item` already uses) or three nullable FKs. Recommend polymorphic — the alternative puts a three-way `COALESCE` in every list query, and a two-table `COALESCE` expression is exactly what drops every index off a filter.

**D3 — Flatten the cost tree or keep it.** The nesting carries real meaning: an outer count ("pick 3 aspects"), a named slot per aspect, and options within a slot. Today every inner select has exactly one option, so a flat `blueprint_costs` row (slot name, resource, min quality, SCU) loses nothing *yet* — but it loses the ability to say "any of these three" the day CIG adds a second option. Recommend storing slot rows plus option rows, and rendering the single-option case as a plain line.

**D4 — How far "where to obtain them" goes.** Three levels, decide which ships first:

1. *Pool membership only* — "this blueprint drops from 3 reward pools", named by their record key. Cheap, and useless to a player.
2. *Pool → contract generator → org* — "Eckhart Security bounty missions, VHRT and up". This is what people actually want, and it needs a **contracts parser that does not exist**, reading `contracts/contractgenerator/**` and its localisation keys.
3. *Wikelo, ORS and XenoThreat called out by name* — 51 blueprints across three special pools that are not generated missions at all.

Recommend shipping 1 in the first pass and 2 as its own phase, with the list filterable by source once 2 lands. Whichever ships, the page must state that 876 of 1608 blueprints have **no** stated source rather than rendering an empty section that reads as a bug.

**D5 — Category names are not in the export.** `blueprintcategorydatabase` lists 21 category refs and not one of them resolves to a record anywhere under `Data/Libs`. So the category GUID can be stored and matched against the fabricator whitelists, but it cannot be labelled. Grouping and filtering must come from the blueprint's folder path (`fpsgear/armour/combat`, `vehiclegear/cooler`) or from the output's own category, which we already have. Recommend the output's category — it is the one a visitor already knows from the ships and components pages.

**D6 — Quality is the point, and it is the hard part.** Every slot ramps a named stat linearly over quality 0→1000 (`gpp_armor_damagemitigation` and 23 others in `craftedproperties/`). A blueprint page that lists only "0.2 SCU of Tungsten" answers none of the question a crafter is actually asking. Decide whether the first pass shows the ramp (a min/max per stat, and the `minQuality` gate per resource) or defers it — but not by accident.

**D7 — Build facts, like every other catalogue.** `blueprints` + `blueprint_builds`, a `current_version` scope, `retire_absent` at the end of the loader run, `prune_builds`. 4.9.0 had 1598 blueprints and 4.10.0 has 1608, so drift is already real and a plain upsert would keep retired recipes visible forever. The commodities, components and equipment loaders all carry this pair already.

**D8 — Where it lives, and #4989.** #4989 asks for a Components surface, and this asks for a Blueprints one. Both are public catalogues of sc_data we currently expose only through the API and admin. Decide the shell once — a shared catalogue section the two hang off, or two independent top-level routes — before either builds its own navigation.

**D9 — Feature flag.** New public surface, so a `blueprints` entry in `config/feature_flags.yml` and the schema regeneration that comes with it.

## Acceptance Criteria

- [x] A `ScData::Parser::BlueprintsParser` writes one JSON per blueprint: output ref, category ref, craft time, the cost tree with slot names, resource refs, min quality and SCU quantities, and the per-slot property modifiers
- [x] A `ScData::Loader::BlueprintsLoader` upserts on the blueprint's game ref, links the output to its `Component` / `Equipment` / `Commodity`, resolves each cost to a `Commodity`, writes build facts and ends with `retire_absent`
- [x] Both are registered in `BaseParser.all` / `BaseLoader.all` and covered by parser and loader tests against the pushed parsed tree
- [ ] A `/catalogue/blueprints` list: search by name, filter by output kind, by category, by required resource, and by whether a source is known; sortable, paginated
- [ ] A blueprint detail page: what it makes (linked to that catalogue entry), craft time, every slot with its resource, min quality and quantity, and the stats each slot moves
- [ ] The reverse link — a component, equipment or commodity page says which blueprints make it, and which blueprints consume it
- [ ] Blueprints with no known source say so explicitly
- [x] Public API endpoints and filters, in the schema, matching the commodities/components/equipment shape — #5013
- [x] Everything behind a feature flag — shipped behind `blueprints`, dropped again in #5017 once the API was proven

## Notes

Data traps found while measuring, all of which the loader has to survive rather than raise on:

- **Three outputs have two blueprints each** — the 890 Jump and Polaris S04 shields, two S01 quantum drives, two S02 powerplants. Almost certainly copy-paste in the game data, but a unique index on the output would fail the load.
- **One output does not exist** — `bp_craft_cool_s04_cnou_pioneer.xml` points at an entity class present in no file in the export.
- **Resource refs do not join on `sc_ref`.** The blueprint names a `ResourceType` GUID; `Commodity#sc_ref` is populated from the *crate* entity and is `null` for 100 of 232 commodities, Iron included. The parser has to resolve the `ResourceType` record to its `@items_commodities_*` key and join on `sc_key`.
- A parser change means a full re-parse and push before loader tests see anything — the loader tests read the pushed tree, not a local one.

---

## Decisions taken (2026-09-17)

Resolved with the maintainer before code. Exec-plan: `docs/exec-plans/4988-blueprints-crafting-catalogue.md` on `feat/4988-blueprints-crafting-catalogue`.

| | decision | note |
|---|---|---|
| **D1** | Name, icon, manufacturer and slug all derive from the crafted output | The output link is mandatory. The one blueprint whose output does not resolve (`bp_craft_cool_s04_cnou_pioneer`) is skipped and counted, not raised on |
| **D2** | Polymorphic `craftable` | No unique index on it — three outputs have two blueprints each |
| **D3** | `blueprint_cost_slots` + `blueprint_cost_options` | The single-option case renders as a plain line. `optionalCost` is ignored, not modelled |
| **D4** | **Full chain to the org, in the first pass** | See the verification below — this changes the plan |
| **D5** | Category comes from the output, not the export | The 21 category refs resolve to nothing; the GUID is stored for the fabricator whitelists but cannot be labelled |
| **D6** | **Ship the quality ramps** | Min/max per stat per slot, plus the per-resource `minQuality` gate. Deferring means a full re-parse later |
| **D7** | `blueprints` + `blueprint_builds`, `current_version`, `retire_absent`, `prune_builds` | As commodities, components and equipment already do |
| **D8** | ~~One shared public catalogue shell, built here~~ — **overtaken.** The shell was built by #4989 in #5021; blueprints joins it as the second tenant | See the status section at the top |
| **D9** | A `blueprints` flag in `config/feature_flags.yml` | With the schema regeneration that comes with it |
| **D10** | **Four stacked PRs** | 1 parser+loader · 2 contract sources · 3 public API · 4 shell + pages |

### D4 verification — the chain is GUID-only

The body above describes `blueprint → pool → ContractGenerator → org`. Measured against `4.10.1-live.12660092`, that chain is real but resolves **only by ref**: `grep -rl "BlueprintPool\|blueprintreward" contracts/` returns **zero** files.

```
blueprint ──> BlueprintPoolRecord.blueprintRewards[].blueprintRecord   (weighted)
BlueprintPoolRecord.__ref ──> ContractGenerator                        (by GUID)
ContractGenerator.factionReputation ──> FactionReputation.displayName
```

Worked end to end:

| step | value |
|---|---|
| pool | `bp_missionreward_foxwellenforcement_ambush.xml` |
| its `__ref` | `af21edb5-3938-45ac-8a35-130e52463f44` |
| referenced by | `contracts/contractgenerator/mercenary_guild/foxwellenforcement/shipbattles/foxwellenforcement_ambush.xml` |
| its `factionReputation` | `016c2950-0292-468c-a408-a5283b52f158` |
| resolves to | `factions/factionreputation/factionreputation_lawful_foxwellenforcement.xml`, `displayName="@Foxwell_RepUI_Name"` |

Counts: **154** pool records, **130** reachable from a contract generator, **107** generator files across **10** guild folders. The 24 unreachable pools are the special ones (Wikelo, ORS, XenoThreat) — they keep their pool name and get no org. Generators carry a localised title per difficulty tier (`@Foxwell_ShipAmbush_{VE,E,M,H,VH,S}_title_001`), which is what makes "Eckhart Security bounty missions, VHRT and up" expressible rather than just "Eckhart Security".

This pulls a `ScData::Parser::ContractsParser` and faction-name resolution into scope as Phase 2. Factions are read through for `displayName` only — no factions table in this pass.

---

## Phase 5 — counted commodities (deferred out of Phase 4)

Phase 4 shows what the reader already holds beside each recipe slot, matched on
the commodity the stock position points at. It cannot say whether a holding is
*enough* for 11 of the 37 crafting commodities, because the two sides measure
in different units and nothing we store can convert them.

### The gap

A recipe states a slot's cost one of two ways, and the parser already keeps them
apart as `cost_type`:

| cost kind | quantity read from | measured in |
|---|---|---|
| `CraftingCost_Resource` | `quantity.SStandardCargoUnit.standardCargoUnits` | SCU |
| `CraftingCost_Item` | a bare `quantity` | pieces |

An inventory cannot express the second: `InventoryLedgerEntry::UNITS_BY_CATEGORY`
maps `"commodity" => ["scu"]`, so a hand- or vehicle-mineable gem recorded as
210 SCU can never be compared with a recipe asking for 170 pieces. Phase 4
therefore shows the row without a sufficiency claim rather than inventing one.

### It is a property of the commodity, not of the recipe

Measured over all 4,289 cost options in `4.10.1-live.12660092`: **26 commodities
are always bulk, 11 are always counted, and none is used both ways.** So the unit
belongs to the commodity and can be stored on it.

`commodity_type` does not predict it — minerals split 9 counted / 10 bulk — so
the crafting data and the entity are the only sources of truth.

### The game files do say so

The commodity's own entity carries it, four ways over:

| | counted | bulk |
|---|---|---|
| `ResourceContainer generateRandomQuality` | `1` | `0` |
| `ResourceContainer` capacity | `SMicroCargoUnit` | `SCentiCargoUnit` / `SStandardCargoUnit` |
| `mutabilityLevel` | `ReadOnly` | `ReadWrite` |
| entity path | `entities/scitem/carryables/1h/harvestable_mineral_1h_*` | `entities/scitem/carryables/2h/carryable_2h_*_commodity_*` |
| interactions | — | `OpenExternalInventory` — it is a crate you open |

Of the 31 crafting commodities whose entity matches by purchasable key, every one
with `generateRandomQuality="1"` is counted (11/11) and every one without is bulk
(20/20) — no crossover. The 6 unmatched are bulk by arithmetic (26 bulk total, 20
matched); their crates carry no `SCItemPurchasableParams` commodity key and are
named from the resource-type database instead.

`generateRandomQuality` is the field to read, and it is the meaningful one: the
counted materials are exactly those whose individual pieces each roll their own
quality, which is why crafting counts them and why the quality ramp applies
per piece at all.

### The 11

Hadanite · Dolivine · Aphorite · Sadaryx · Beradom · Glacosite · Feynmaline ·
Janalite · Carinite · Saldynium (Ore) · Yormandi Eye

### Scope

1. `ScData::Parser::CommoditiesParser` reads `ResourceContainer` — it currently
   reads `SCItemPurchasableParams` and the resource-type database only.
2. A boolean on `commodities` (build-fact layered, as the other columns are).
3. `UNITS_BY_CATEGORY` offers `units` alongside `scu` for a counted commodity.
   Widen rather than flip: every existing row is SCU and stays valid.
4. The blueprint slot's capacity check becomes exact instead of skipped.

### Notes before starting

- Needs a re-parse to land, which is its own step — see the parser/re-parse note
  in the discovery log.
- Inventory positions are keyed by `name`/`category`/`unit`, so the same
  commodity held in both units becomes two positions.
- Touches the shipped inventory feature — unit picker, transfers, stock
  aggregation — not just the catalogue.

🤖

mortik:
Phase 4 of #4988: the public pages, and the API the pages needed.

Phases 1–3 landed the parser, the loader, the contract sources and the read API. This is what a reader actually sees.

## The pages

**`/catalogue/blueprints/`** — the list, joining components under a new Catalogue nav section. Rows carry the materials a recipe consumes, each one a filter link. Filters: name, what it makes, material (multi-select over the 37 the recipes actually ask for, not the 232-row commodity catalogue), and whether anything hands the recipe out.

**`/catalogue/blueprints/:slug/`** — the recipe. Each slot carries its own grade, because a crafter buys each material separately and one page-wide figure would assert they are the same. The page opens at 500.

## What better material buys, as a figure

A modifier states a factor and says nothing about what it is a factor of, so `1.05×` was the whole answer where "damage 250 → 262.5" was the question. `BlueprintStatBases` matches the stat a modifier names to the figure the catalogue already holds for the crafted item.

It resolves **3,799 of the 5,859 valued modifiers (65%)**. The rest — component health, weapon recoil, quantum fuel burn, hull scraping — have no counterpart here and keep showing their factor rather than a number we invented.

That the catalogue's figure is the right one to multiply is measured, not assumed: **3,933 of the 3,963 single-segment ramps are exactly symmetric about 1.0**, and every piecewise stat seams at the 500/501 boundary. Quality 500 is the grade the catalogue describes, which is also why the page opens there.

A "What you get" panel composes the slots: **358 of 1,142 recipes move one stat from two different slots**, and neither slot's own panel can state the result.

## What you already hold

Signed-in readers with inventories see their own stock beside each slot, with the held grades marked on the quality track. Clicking a mark or its row sets the slot; clicking it again returns to neutral.

Matching is on the record a stock position points at, never on its name — a position is named by whoever wrote it down, so "iron ore" against the commodity "Iron" would both miss and mis-match. That needed the aggregate stock endpoints to stop dropping the link, which is the first commit.

Own stock and fleet stock are listed apart: a fleet's store is not the reader's to spend. A row shows only where it could actually fill the slot — except for the eleven commodities the game counts in pieces while an inventory can only record SCU, where no comparison is possible and none is claimed. That gap is written up as phase 5 on the issue.

## Smaller things

- **`RowList`** extracted behind both catalogues — written out twice they had already drifted.
- **Components detail** names the recipe it is crafted from (476 of the catalogue can be made).
- Mission titles render their unresolved placeholders as what varies — **1,293 of 4,030** carry one.
- Two contract bugs neither type system could catch: a `decimal` column rendering as a JSON string while the schema called it a number, and `route.query` handing back a bare string for a param the schema declared an array. Both have regression tests.

## Checks

2,997 Ruby tests · 1,897 frontend tests · `lint:ts`, `rubocop` and the production build clean. The slot card is in visual-tests under its own Blueprints page, with the no-figures and below-the-gate cases.

Resolves #4988 (phases 1–4; phase 5 is recorded on the issue).

🤖

## Summary by CodeRabbit

* **New Features**
  * Added a searchable, filterable Blueprints catalogue with sorting, pagination, material filters, and detailed blueprint pages.
  * Blueprint details now show materials, crafting outputs, quality controls, stat previews, inventory availability, crafting sources, and craft times.
  * Added catalogue navigation for Components and Blueprints, plus “crafted from” links on component pages.
  * Inventory views now identify matching catalogue items and vehicle locations where available.
* **API Updates**
  * Added blueprint material filters and expanded blueprint, modifier, and inventory response details.
  * Added support for filtering blueprints by multiple consuming materials.
* **Localization**
  * Added blueprint and catalogue translations across supported languages.


mortik:
The 3/5 above is against `5d0587ee5` — its own footer says so — and all three findings it lists were fixed in `3d0d6dd80`:

- **False filter dropped** — right, and the cause was `useFilters` deleting every falsy value before it builds the route, so boolean `false` never reached the URL. The option keeps the string it carries (which is all a query string can hold anyway) and the API casts it, as it already did. Confirmed against the live endpoint: 706 with a source, 901 without, 1607 unfiltered.
- **Material IDs missing** — right, `id` is required by `BlueprintCostCommodity`. Added.
- **Item schema inline** — right. Now `InventoryStockItemRef`. Deliberately *not* `InventoryItemRef`: that one already exists and means a ledger entry's link, which always resolves and carries the record's name and availability. I did overwrite it by mistake first and oasdiff caught it as 11 `response-property-became-nullable` errors; it is restored byte-identical to main.

Also from review, in `eb67e7680`:

- **Unanimous, non-null catalogue identity** — right, and it mattered more than the contract: `COUNT(DISTINCT)` skips nulls, so one linked entry beside one free-text entry claimed the link *and* the position's whole quantity. Both aggregates now require every entry to name the same record, with a test each.
- **Stock caches survive logout** — right, and specific to this PR: the stock query is disabled when signed out rather than absent, and the catalogue is public, so the next person to open a recipe in that tab would have seen the previous reader's hangar. `clearSession` already did this for the supporter claim key for the same reason.
- **Slider focus indicator** — right, the outline was dropped with nothing in its place. The thumb carries it now.
- **Stale disclosure keys** — right, a page reused for another recipe kept the old keys and left every group shut.

**Not taken:** the stylelint findings. There is no stylelint config or script in this repo and no check runs it, and the quoted `"Orbitron"` they object to is the form `metricsCard.scss` and the hardpoints components already use — matching it was deliberate. Happy to change it if we want to adopt stylelint, but that is its own change.

On the breaking-change gate: widening `consumingCommodity` to accept a list reads to oasdiff as *narrowing* the type (`string` → `string, array<string>`), as `anyOf` or as a union. So it is a second parameter now, named the way this schema already names the pair — `craftableTypeEq` beside `craftableTypeIn`. Verified locally with oasdiff against main: no breaking changes on either schema.

🤖

Issue #5002, opened 2026-09-17
sc_data parser double-encodes a component's tags


mortik:
## Description

`ItemsParser` writes a component's `tags` as **the whole JSON array encoded into a single
string, inside an array** — so every component in the parsed tree carries a value shaped like:

```ruby
["[\"flightReady\"]"]     # what we store
["flightReady"]           # what it should be
```

`required_tags` on the same record is correct (`["drak_ironclad"]`), which is what makes this
look like a parser bug on one field rather than a format question.

## Scale

Every component with tags is affected. Sampled 4,000 rows with a non-empty `tags`: **4,000 of
4,000** are double-encoded. The parsed tree itself already holds the bad shape — e.g.
`data/sc_data/parsed/live/items/a18_door_ionorth_entrance.json` has `tags: ["[]"]` beside
`required_tags: []` — so this is upstream of the loader, which passes `item["tags"]` straight
through (`app/lib/sc_data/loader/items_loader.rb:24`).

## Why it matters now

`tags` and `required_tags` are together the answer to "what fits where" — a port and the item
that fits it name a shared tag. That is exactly the question a Components catalogue visitor
asks, so the public payload wants both.

`required_tags` ships in the components catalogue work (#4989). **`tags` is deliberately held
back** from the public payload, because exposing it today would publish visibly broken values.
It can be added in the same place the moment this is fixed.

## What it needs

- The `tags` extraction in `ScData::Parser::ItemsParser` corrected to store the parsed array.
- A full re-parse and push of the tree — loader tests read the pushed tree, not a local one.
- No migration: the loader rewrites `tags` on the next run, and the column is already an array.

Worth doing alongside #4997, which is also a parser fix needing a re-parse.

🤖

mortik:
Checked against #5001, which landed in the meantime: it does **not** fix this. That change swaps `translate` for `localize_name` on `name` and `short_name` only — the `tags` extraction is untouched, and the parsed tree still holds `tags: ["[]"]` beside a correct `required_tags: []`.

So this stays open and independent. `requiredTags` ships in #5003; `tags` joins it the moment this is fixed. 🤖

Issue #4989, opened 2026-09-17
Components — a public catalogue with per-category metrics


mortik:
## Status — done, 2026-09-21

**All five phases have shipped**, and the three follow-ups this issue was carrying — the history
tab, the facet labels and `tags` in the payload — are merged. Plan:
`docs/exec-plans/4989-components-catalogue.md`.

| phase | PR | state |
|---|---|---|
| 1 — a unique slug, `type_data` to `jsonb` | #5000 | **merged** |
| 2 — the public API (`show`, real sorting) | #5003 | **merged** |
| 3 — the per-category metric renderer | #5006 | **merged** |
| 4a — a page per component, and the link from every hardpoint | #5015 | **merged** |
| 4b — the catalogue list, and `/catalogue` as the section entry | #5021 | **merged** |
| 5 — UEX component prices | #5084, #5097 | **merged** |

Also merged on the way: **#5016** dropped the `components` feature flag, **#5020** turned each of
the eight metrics into a `Gteq`/`Lteq` filter rather than only a sort, **#5022** serves the last
loaded build while a newly configured one waits for its import, **#5007** fixed the parser
writing `tags` as the array's own inspect output, and **#5025** put a sort line on the card lists
that had no way to sort.

### The remaining work, as sub-issues — all closed

| | issue | closed by |
|---|---|---|
| UEX component prices (Phase 5) | #5070 | #5097, on the syncer from #5084 |
| The history tab (D10) | #5071 | #5075 |
| `tags` in the payload | #5072 | #5096 |

### #5084 and #5097 — Phase 5: what a component costs, and what silence means

- **`Uex::PriceSyncer` only ever priced models.** `ITEM_TYPE = "Model"`, and `item_prices` held no
  `Component` row, so `availability` was two empty arrays on every component the API has ever
  served. `Uex::ComponentPriceSyncer` fills it — no migration, because `ItemPrice#item` is already
  polymorphic and `Component` already includes `ItemPriceConcern`.
- **The matcher refuses more than it could.** There is no `uex_id` column to join on, so every run
  resolves from scratch: hand-written `MAPPINGS`, then `sc_ref` → UEX `item_uuid`, then an exact
  name **only when exactly one component answers to it**. A name several components share is left
  unpriced and reported — the game files carry a row per mount, and putting a shop price on a
  gimbal welded to one ship is worse than saying nothing. 519 components priced, 4,193 rows across
  111 locations, measured against the live catalogue.
[...]

PR #5096, opened 2026-09-20
feat(components): publish the tags an item carries

### LINKED ISSUE #5072: Components — tags in the payload, after a re-parse
Part of #4989. `tags` is the one field the acceptance criteria name that the payload still does
not carry.

## Where it stands

The **parser is already fixed** — 33f60236b (#5007), "keep each of an item's tags, not the
array's inspect output". `normalize_tags` in `app/lib/sc_data/parser/base_parser.rb` splits the
raw value and strips the `$` that marks a tag as taking part in port matching.

What is left is the **data**. The pushed parsed tree was built before that fix, so the values in
it are still the broken ones: the whole JSON array as a single string inside the array
(`["[\"flightReady\"]"]`). `app/views/api/v1/components/_base.jbuilder` leaves `tags` out on
purpose and says so, because exposing it would publish visibly broken values.

`requiredTags` already ships — it goes through the same helper and reads correctly.

## Why this matters

`tags` and `requiredTags` together are what decide which port an item fits, which is exactly the
question a catalogue visitor has. Half of that answer is missing.

## Steps

1. Re-parse the tree so the pushed values carry real tags, and load it.
2. Add `tags` to `_base.jbuilder` beside `required_tags`.
3. The schema component under `app/api_components/`, then `./bin/generate-schema` and an orval
   regeneration.
4. Render it on the detail page next to `requiredTags`, which is already at
   `pages/components/[slug].vue:100`.

## In the way

- **A re-parse of the same version is invisible.** CI's cache and the nightly loader both key on
  the version, so re-parsing without a bump changes nothing anyone can see — bump the parser
  version and load by hand.
- **The payload is cached.** `_component.jbuilder` caches on a key that a new field does not
  change, so the field stays invisible until the key moves.
- `inventoryConsumption`, `ammunition`, `powerConnection` and `heatConnection` stay out
  regardless. They are raw game-file dumps — `heat_connection` alone carries 22 different keys
  across the table, mixing `MaxCoolingRate` with `cooling_rate` — so no honest schema can
  describe them until the parser gives them consistent keys.

🤖

mortik:
Closes #5072. Part of #4989 — one of the two things that issue still carried.

`tags` and `requiredTags` together answer "what fits where": a port and the item that goes
into it match by naming the same tag. Only half of that was published, and the missing half
is the item's own.

It was held back because the values were broken, not because the question was uninteresting.
The parser wrote the whole JSON array as a single string inside the array, so every component
in the tree read `["[\"flightReady\"]"]` — publishing that would have been publishing a bug.

## The data was already fixed

Two things had to be true before the field could ship, and both already are:

- **The parser fix landed in #5007** (`normalize_tags` splits the raw value and drops the `$`
  that marks a tag as taking part in port matching).
- **The pushed tree is the re-parsed one.** `bin/scdata cache-key` reads `version.json` out of
  the bucket, and its live checksum is `128ffc21…52db82` — byte for byte what the local tree
  carries. 3,413 of the 7,795 items in it have tags, and none of them is a re-encoded array.

So there is nothing to push and no cache prefix to bump: since #5076 both CI and `CheckJob`
key on the tree's own checksum rather than on the build version.

Production is still on `4.10.0-live.12519617` — three days behind `main`, which has pointed at
`4.10.1-live.12660092` since #4991. The next deploy moves it, and the loader run that follows
writes the tags this serves. Nothing here depends on that having happened first: the field is
absent from the payload today, so it cannot get worse.

## What changed

| | |
|---|---|
| `_base.jbuilder` | `json.tags` beside `required_tags` |
| `_component.jbuilder` | cache key `v2` → `v3` |
| `Shared::V1::Schemas::Component` | `tags: {type: :array, items: {type: :string}}` |
| detail page | a row beside "Fits a port offering", in all seven locales |

The cache key is not optional. A cached fragment *is* a rendered payload, so a new field is
invisible on every component that has been served once until the key moves.

## The loader test reads the real tree

Deliberately, and it is the second test in that file to do so. The risk this guards is not a
loader that stopped reading the field — it is a tree parsed before #5007, which no fixture can
have. It asserts that thousands of components arrive with tags and that none of them carries a
value starting `[`.

The fixture tree carries no tags at all, so a fixture-based test would have asserted nothing.

## Not in here

`inventoryConsumption`, `ammunition`, `powerConnection` and `heatConnection` stay out, for the
reason the jbuilder comment already gives: they are raw game-file dumps — `heat_connection`
alone carries 22 different keys across the table — and no honest schema can describe them until
the parser gives them consistent keys. That is #5002's cleanup, and the comment now points at
it rather than at `tags`.

## Checks

- `test/integration/api/v1/components_show_test.rb` — 4 tests, green
- `test/loaders/sc_data/items_loader_test.rb -n "/one entry per tag/"` — green against the real
  tree
- `standardrb` clean; `pnpm lint:ts` exit 0
- Schema regenerated. The `components.parameters` block reordered itself the way it does on
  macOS, so the generated files carry only the four added lines each.

🤖

## Summary by CodeRabbit

* **New Features**
  * Component details now display assigned tags when available.
  * Component API responses include tags as an array of values.
  * Component schemas document the optional tags field across supported APIs.
  * Added localized tags labels for supported languages.

* **Bug Fixes**
  * Tags are now provided as individual values instead of a serialized array string.


mortik:
Fixed the one finding: the mangled-tag check covered only the first 500 of the ~3,400 tagged components, so a partial regression could have slipped past it.

It now asks the whole column at once instead of a sample — a tag starting with `[` is stored as the two characters `"[`, so a `LIKE` finds every one of them in SQL rather than deserialising 3,400 rows. Complete coverage, and it costs less than the batched loop did (252s against 296s for `find_each`, on a test that ran 215s before either).

Green locally against the real tree. 🤖

